PR #21, opened 2026-07-12
Fix val-metric statefulness, HParams expansion, per-image PSNR, config validation (P2.2, P2.3, P2.4, P2.6)

YeapJieShen:
Compute val PSNR/SSIM via torchmetrics.functional to remove dead accumulator state (P2.2), validate psnr_channels (P2.3), expand nested dataclass configs into individual HParams via asdict (P2.4), and make val PSNR a per-image mean invariant to batch size (P2.6).

**Stacked PR** - base: `stack/04-pr4` (merge after its base). Part of the triage-delivery stack of 11 PRs; the full stack is 216 tests passing and ruff-clean.

See triage items in the title.

> Generated with Claude Code

### DIFF
--- a/sisr/training/config.py
+++ b/sisr/training/config.py
@@ -84,3 +84,19 @@ class SREvalConfig:
     crop_border: int = 0
     psnr_channels: list[str] = field(default_factory=lambda: ['RGB'])
     separate_psnr: bool = False
+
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, ready to merge.

PR #28, opened 2026-07-29
Adopt conda + uv for local development, move test CI to uv

YeapJieShen:
Retires the in-project `.venv` in favour of a named conda env (`sisr`, Python 3.13), installs packages with uv from inside that env, and moves the test workflow to uv.

## What changed

- **CI** — `test.yml` installs via `uv pip install --system`, with the uv cache scoped per matrix leg. `build.yml` deliberately stays on plain `pip install .`, so the end-user pip path remains verified on every PR across all four matrix legs. Gate job names are unchanged, so required-check contexts do not move.
- **VS Code** — adds `python.testing.pytestArgs: ["-n", "0"]` so Test Explorer breakpoints and per-test results work despite the `-n 8 --dist=loadscope` in pyproject's `addopts`. Terminal `pytest` keeps all 8 workers. No interpreter path is committed; it is selected via the palette and stored in VS Code's own workspace storage.
- **Docs** — conda + uv setup in CONTRIBUTING, uv variants in the README, with pip retained as the supported path throughout.

## Notes

uv is documented as `python -m uv`, not bare `uv`: uv resolves `VIRTUAL_ENV` ahead of the active conda env, so the bare binary installs into whatever virtualenv happens to be active. A bash-style `--python "$CONDA_PREFIX"` guard was rejected because it expands to the empty string in PowerShell.

`astral-sh/setup-uv` is pinned to the exact patch `v9.0.0` rather than a moving major tag — the action stopped publishing moving tags at v8.0.0 as supply-chain hardening.

Two doc corrections ride along: CONTRIBUTING's `[dev]` extra list omitted `pytest-xdist`, `pyyaml` and `ruff`, and it claimed the `build` check exercises `pip install -e ".[dev]"` when that job runs `pip install .`.

No dependency or version constraint changed. No `uv.lock`, no `requirements.txt`, no `[tool.uv]` block.

## Verification

Locally in the `sisr` env: `torch 2.13.0+cu132` with `cuda.is_available() == True`, 216 tests passed, `ruff check` and `ruff format --check` clean.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,13 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
+      # setup-uv stopped publishing moving major/minor tags at v8.0.0 as supply-chain
+      # hardening, so this pins an exact patch unlike the @v4/@v5 actions above.
+      - uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
       - name: Install ruff
-        run: pip install ruff
+        run: uv pip install --system ruff
       - name: Ruff lint
         run: ruff check --output-format=github .
       - name: Ruff format check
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #32, opened 2026-07-31
Add construction-time config validation seam (INIT.16)

YeapJieShen:
Fail-fast validation for template fields that must agree but are written in **different forms**, so YAML anchors cannot express the relationship.

Anchors (#31) solve *"the same value repeated"*. They cannot solve *"different values that must agree"* — e.g. `num_channels: 1` ↔ `YChannelProcessor` ↔ `example_input_shape: [1, ...]`, three encodings of "we work in Y". Setting `num_channels: 3` with `YChannelProcessor` silently trains the wrong thing.

### Design

**Template method, not `@abstractmethod`.** Both config bases are concrete and **directly instantiated as fallbacks** (`training_config or SRTrainingConfig()`), so making either an ABC would break that path.

| Layer | Responsibility |
|---|---|
| `SRTrainingConfig.validate_against(model, processor)` | Universal: `example_input_shape[0] == processor.model_channels`, plus a `torch.no_grad()` forward probe |
| `SRCNNTrainingConfig` / `SRResNetTrainingConfig` | Own channel field first (readable error), then `super()` |
| `SRLightning.__init__` | Orchestration only — **zero architecture knowledge** |

The forward probe validates the *actual* `nn.Module`, so it **cannot drift** — it is not a description of the model, it is the model. Subclass checks exist only to give better errors than a raw shape mismatch. Guarded on `example_input_shape` being set, since it is optional.

New abstract `SRProcessor.model_channels` — RGB 3, Y 1, YCbCr 3. A fact only the processor knows, not a mirror of the model.

### Also: `SREvalConfig.psnr_keys`

Moves PSNR key derivation off `SRLightning.__init__` onto the config that owns `psnr_channels` / `separate_psnr`. Verified byte-for-byte equal to the old inlined loop for both `separate_psnr` settings. This public seam is what lets the benchmark logger and validation stop disagreeing (follow-up PR).

### Rejected alternative

A parallel pydantic schema — it would be a *third* validation system (jsonargparse already type-validates real `__init__` signatures), it cannot see across the seam anyway since the correlations span *sibling* objects, and a schema *describes* rather than *tests*, so it goes stale when `__init__` changes.

Follows the existing precedent: `SREvalConfig.__post_init__` already validates `psnr_channels`.

**Tests:** 217 → **243** (26 new). `ruff check` / `ruff format --check` clean.

### DIFF
--- a/sisr/models/srcnn/config.py
+++ b/sisr/models/srcnn/config.py
@@ -12,6 +12,8 @@ Reference: Image Super-Resolution Using Deep Convolutional Networks
 from dataclasses import dataclass, field
 from typing import Literal
 
+from sisr.models.base import SRModel
+from sisr.processors.base import SRProcessor
 from sisr.training.config import SREvalConfig, SRTrainingConfig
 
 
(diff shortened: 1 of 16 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #33, opened 2026-07-31
Fix benchmark PSNR key over-emission; harden checkpoint filenames (P2.7, P5.9)

YeapJieShen:
> **Stacked on #32** (`feat/config-validation-seam`) — merge that first. This PR targets that branch, so the diff shows only its own changes.

### P2.7 — `separate_psnr` was silently inert on the benchmark path

`BenchmarkImageLogger._collect_batch` built `psnr_dict` by iterating **every** tensor `_build_psnr_tensors()` returns — and that method populates all four of `RGB/R/G/B` if *any* RGB-family key is requested (likewise `YCbCr/Y/Cb/Cr`). So a config asking for 2 keys emitted **8**.

Observed in a real run's `metrics.csv`:

```
Set5_psnr(B),Set5_psnr(Cb),Set5_psnr(Cr),Set5_psnr(G),Set5_psnr(R),Set5_psnr(RGB),Set5_psnr(Y),Set5_psnr(YCbCr),Set5_ssim
```

alongside only `val_psnr(RGB),val_psnr(YCbCr),val_ssim` from validation.

Now iterates the public `eval_config.psnr_keys` (#32's seam). This also shrinks `_flush_buffer`'s output, which derives its key set downstream — so per-set means and **per-image TensorBoard tags** shrink too. Set5 previously emitted 40 per-image tags per val run where 10 were configured.

### P5.9 — reduced private coupling

Key *selection* no longer reaches into `SRLightning` privates. `_build_psnr_tensors` is still called as a **value lookup**, with an inline note explaining why: the colorspace split has no public seam, and duplicating it in the callback would recreate the exact divergence P2.1 removed.

### Checkpoint filename hardening

`_format_checkpoint_name` does **no** sanitisation — with `auto_insert_metric_name=True` it interpolates the raw metric name, so a `/`-bearing monitor yields `sr-step=1000-psnr/val/RGB=30.1234.ckpt`. It doesn't crash: `TorchCheckpointIO.save_checkpoint` calls `fs.makedirs(os.path.dirname(path), exist_ok=True)`, so Lightning **silently creates a `psnr/val/` directory tree per save**.

Fixed with `auto_insert_metric_name=False` plus a self-supplied `label = monitor_metric.replace("/", "_")`. The monitor still resolves against the real tag — a `/` in a dict key is fine; only the *printed* label is flattened. Landing now because the queued TensorBoard tag-hierarchy PR depends on it.

### Monitor validation moved earlier

Lightning already raises `MisconfigurationException` for a missing monitor — but only after the val loop has run, which under the templates' `val_check_interval: 20000` is a **20k-step-late** crash. `SRCheckpoint.setup` now checks at startup. This moves an existing failure earlier rather than adding a new guard, and is scoped to the `fit` stage so it can't reject configs valid for `validate`/`test`/`predict`.

**Tests:** 243 → **251**. Reviewed independently: spec PASS, quality APPROVED, two Minor findings both fixed in `e5ec1ea`.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -293,6 +293,10 @@ class BenchmarkImageLogger(Callback):
             if n > 0:
                 sr_4d = sr_4d[..., n:-n, n:-n]
                 hr_4d = hr_4d[..., n:-n, n:-n]
+            # Still a private reach (P5.9): the colorspace split has no public
+            # seam, and duplicating it here would recreate the divergence P2.1
+            # removed. Keys now come from eval_config, so this is a value
+            # lookup only — the callback no longer decides *which* keys exist.
             psnr_tensors = pl_module._build_psnr_tensors(sr_4d, hr_4d)
             psnr_dict = {
                 key: torchmetrics.functional.image.peak_signal_noise_ratio(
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #34, opened 2026-07-31
Add LR-only prediction path: predict_dataloader, predict_step, SRPredictionWriter (P3.7)

YeapJieShen:
> **Stacked on #32** (`feat/config-validation-seam`) — merge that first. This PR targets that branch, so the diff shows only its own changes.

### The gap

`sisr predict` was entirely non-functional. `SRDataModule` had **no `predict_dataloader`**, and `SRLightning` did not override `predict_step` — so Lightning's default would call `self(batch)` on the datasets' `(lr, hr)` tuple and crash.

The deeper problem: **no public path took an LR image to an SR image.**
- `forward()` runs the model *without* the processor — for SRCNN it expects a Y-channel tensor, not an image.
- `predict_rgb(lr, hr)` requires an HR reference that doesn't exist at inference.

Both existing entry points assumed a paired evaluation context. This is the first genuine inference path.

### What makes it work

`YChannelProcessor.reconstruct(sr, lr_rgb)` needs the LR RGB back for chroma — and at inference the LR RGB **is** available. So the full `extract → model → reconstruct` pipeline runs end-to-end without HR; only HR-dependent cropping and scoring drop away.

### `_forward_sr` was extracted, not forked

`_forward_sr` is documented as *"the single source of truth for the forward pipeline … so the training and benchmark-logging paths cannot silently diverge."* Rather than duplicating it, the extract→model→reconstruct core moved into a new `_forward_lr`, which `_forward_sr` now delegates to before applying its unchanged centre crop. `predict_step` calls `_forward_lr` directly.

Independently reviewed and traced line-by-line: `_step`'s behaviour is **unchanged** — same call sites, same arguments, same order, zero logic duplicated.

### Also

- `PredictDataset` reuses `SRDataset` for discovery and the uint8→float32 adapter — no reimplemented globbing.
- `SRPredictionWriter` (a `BasePredictionWriter`) saves 8-bit PNGs to a configurable dir, clamping out-of-range output.
- **Duplicate filename stems now raise at construction** — outputs are named by stem, so `cat.png` + `cat.jpg` would have silently overwritten each other (review finding, `5f7e7db`).

### Tests

243 → **265**. End-to-end `Trainer.predict(...)` for **both** architectures, which exercise genuinely different processor paths: SRCNN through `YChannelProcessor` (non-trivial chroma stitching, same-resolution 36×36) and SRResNet through `RGBProcessor` (identity reconstruct, ×2 upsampling to 72×72).

### Known scope boundaries

- `train_dataset` / `val_dataset` remain required on `SRDataModule` even for predict-only runs.
- `predict_dataset` is typed `dict[str, Any] | None`, matching the existing `train_dataset`/`val_dataset` convention where each spec carries its own `class_path`. The CLI therefore takes it as a JSON spec — consistent with the existing design, not a new wart.
- Templates deliberately untouched.

Reviewed independently: spec PASS, quality APPROVED, one Minor finding fixed.

### DIFF
--- a/sisr/datasets/predict.py
+++ b/sisr/datasets/predict.py
@@ -27,12 +27,24 @@ class PredictDataset(SRDataset):
             inference on.
 
     Raises:
-        ValueError: If no image files are found in ``img_dir``.
+        ValueError: If no image files are found in ``img_dir``, or if two
+            images share a filename stem.
     """
 
     def __init__(self, img_dir: str | Path):
         super().__init__()
         self._index_images(img_dir)
+        # SRPredictionWriter names outputs by stem, so `cat.png` and `cat.jpg`
+        # would silently overwrite each other. Fail at construction instead.
+        stems: dict[str, Path] = {}
+        for path in self.img_paths:
+            if path.stem in stems:
+                raise ValueError(
+                    f"Duplicate filename stem {path.stem!r} in {img_dir}: "
+                    f"{stems[path.stem].name} and {path.name}. Predictions are "
+                    f"named by stem, so these would overwrite each other."
+                )
+            stems[path.stem] = path
         self._to_tensor = self._to_tensor_transform()
 
     def __len__(self) -> int:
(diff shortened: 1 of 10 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #35, opened 2026-07-31
Restructure TensorBoard tags into a metric-first hierarchy (INIT.8)

YeapJieShen:
> **Stacked on #33** (which is stacked on #32). Merge #32 → #33 → this. Also contains a merge of `chore/template-dry` (#31), since the rename must touch templates that PR deduplicated.

### The problem

TensorBoard groups on `/`, and almost nothing had one. `train_loss`, `val_loss`, `val_ssim`, `val_psnr(RGB)`, `grad_norm`, `Set5_psnr(RGB)`, `Set5_ssim` were all top-level. Worse, per-image tags `Set5_psnr(RGB)/{filename}` created a **group** named `Set5_psnr(RGB)` sitting beside a same-named **scalar**.

### The scheme — metric-type first

```
loss/train              loss/val
psnr/val/RGB            psnr/val/YCbCr
psnr/Set5/RGB           psnr/Set14/RGB
ssim/val                ssim/Set5
per_image/Set5/psnr/RGB/{filename}
diag/grad_norm
```

Chosen over split-first grouping (`train/loss`, `val/psnr/RGB`) because metric-kind-first puts train and val loss in one panel for direct comparison and keeps all PSNR together regardless of split.

Per-image tags carry a `{key}` segment the illustrative design omitted — both architectures default to `psnr_channels: ["RGB", "YCbCr"]`, so without it the RGB and YCbCr per-image means collide on one tag.

### Why this needed #33 first

The monitor string is interpolated into the checkpoint filename, and `/` is a path separator. Lightning's `_format_checkpoint_name` does **no** sanitisation, and `TorchCheckpointIO.save_checkpoint` calls `fs.makedirs(os.path.dirname(path), exist_ok=True)` — so a renamed monitor would have **silently created a `psnr/val/` directory tree per save**.

#33 defused that (`auto_insert_metric_name=False` + a self-supplied flattened label). Verified by a real `fit` run here: filenames come out flat as `sr-1-psnr_val_RGB=7.1435.ckpt`, with no nested directories under `checkpoints/`.

### Landed in lockstep — all five, or the monitor silently stops firing

Emitted tags · HParams registration in `on_train_start` · `BenchmarkImageLogger` per-set and per-image tags · `monitor_metric` in both tracked templates · `SRCheckpoint.setup`'s validation set. Reviewer independently grepped for stragglers: none.

### Two independent perf fixes

- **`GradNormLogger`** called `.item()` **per parameter** — one GPU sync each. Now stacks per-parameter norms on-device and takes a single `vector_norm`: one sync instead of N. Same value, since the L2 norm of per-partition norms equals the global L2 norm.
- **`WeightHistogramLogger`** default `100` → `10000`. At the templates' 1M-step schedule the old default meant ~10k histogram writes **per parameter**, and histograms dominate event-file size — multi-GB runs. Never exercised at that scale, since neither tracked template wires this callback.

### Known limitation

`LearningRateMonitor` is a stock Lightning callback naming its own tags (`lr-SGD/pgN`); it stays outside the scheme rather than wrapping Lightning internals to force it.

**Tests:** 251 passing. Reviewed independently: spec PASS (all five lockstep points traced), quality APPROVED, one Minor docstring fix in `0b4dcca`.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -45,9 +45,9 @@ class BenchmarkImageLogger(Callback):
     SR output is center-padded when smaller than HR (``padding='valid'``) so
     all three panels share the same spatial size.
 
-    Per-image PSNR/SSIM scalars go under ``"{name}_psnr/{filename}"`` and
-    ``"{name}_ssim/{filename}"``; per-set means under
-    ``"{name}_psnr({key})"`` and ``"{name}_ssim"``.
+    Per-image PSNR/SSIM scalars go under ``"per_image/{name}/psnr/{key}/{filename}"``
+    and ``"per_image/{name}/ssim/{filename}"``; per-set means under
+    ``"psnr/{name}/{key}"`` and ``"ssim/{name}"``.
 
     Border cropping is sourced from ``pl_module.eval_config.crop_border`` at
     the use site; this avoids dual-knob configuration drift.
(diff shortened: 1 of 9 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #36, opened 2026-07-31
Cache SRResNet HR images raw in LMDB, eliminating per-crop PNG re-decode (INIT.13)

YeapJieShen:
> **Stacked on #31** (`chore/template-dry`) — merge that first.

### The problem, measured

`TrainDataset.__getitem__` decoded a full DIV2K HR PNG to keep one 96×96 crop. Profiled over 20 real images (mean 2.92 MP):

| Stage | Mean | Share |
|---|---|---|
| **decode (full PNG)** | **108.8 ms** | **98.94%** |
| random crop (slice) | 1.04 ms | 0.95% |
| resize 96→24 | 0.12 ms | 0.11% |

**0.316% pixel utilisation, 316× amplification.** And because `__len__` is `len(img_paths) * crops_per_image` while `__getitem__` indexes `img_paths[idx % len(img_paths)]`, `crops_per_image: N` **re-decoded the same image N times**.

This was invisible to every GPU-side optimisation — AMP, `channels_last`, `torch.compile` all leave it untouched. It's also why GPU augmentation couldn't have helped: the op being moved was ~1% of the cost.

### The fix

Cache each HR image **raw** in LMDB; random-crop and derive LR in-process at load. Crop randomness is fully preserved — an oversized-tile compromise (cache 128×128, crop 96×96 within) was rejected, because the point is eliminating decode, not reducing bytes read.

**Measured on the full 800-image DIV2K set:** build 35.3 s one-off; warm steady-state read **~0.05 ms**. Cold first touch is ~11–15 ms (real disk I/O on a fresh ~6.6 GB mmap), paid once per image rather than once per crop.

### Three things SRCNN's cache didn't need

1. **Zero-copy reads.** `LMDBCache.get()` copies the whole value, which would discard most of the win. New `get_buffer()` is a **context manager** yielding the memoryview, so the buffer's lifetime is structurally tied to its transaction rather than to caller discipline.
2. **Per-image shape metadata.** DIV2K images vary, so an 8-byte `struct("<II")` (H, W) header prefixes each value — one lookup, no second key.
3. **`map_size`** computed from real per-image header sizes × 1.1 (not hardcoded), floor 64 MiB.

`crops_per_image: N` now takes N cached reads instead of N decodes.

### Memory safety

LMDB buffers with `buffers=True` are valid **only inside the transaction**, and neither `np.frombuffer` nor slicing copies — so a returned crop could alias freed mmap pages. The dataset `.copy()`s the crop **before** the `with` block exits; nothing stores the buffer or a derived view.

The reviewer verified this empirically rather than by inspection: reproduced the hazard with an uncopied view (bytes turned to garbage after ~1000 write commits), then confirmed the `.copy()` pattern used here survives identical stress.

### Scope

SRCNN's cache is deliberately untouched — it stores both `lr_` and `hr_` keys and converting it is a separate PR. `sisr/cache.py` remains torch-free (guard still passing); build workers run under spawn and would otherwise pay a ~10s torch import each.

**Tests:** 217 → **230**. Reviewed independently: spec PASS, quality APPROVED, one Minor documented in `0cd2fa6`.

### DIFF
--- a/sisr/cache.py
+++ b/sisr/cache.py
@@ -14,8 +14,9 @@ torch-free; colorspace math lives in :mod:`sisr.colorspace` for this reason.
 
 import os
 import shutil
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterator, Sequence
 from concurrent.futures import FIRST_COMPLETED, Future, ProcessPoolExecutor, wait
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #37, opened 2026-07-31
Wire throughput levers into the shipped templates (INIT.12)

YeapJieShen:
> **Stacked on #35.** Merge chain: #32 → #33 → #35 → this.

### What landed

- `matmul_precision: high` now set in both tracked templates. `sisr/cli.py` already implemented the key; no tracked template used it.
- **`precision: bf16-mixed` ships commented out.** This is a paper-reproduction repo, so the shipped default must train in full fp32. Verified inactive by resolving both templates through the real CLI (`trainer.precision=None`).
- `cudnn_benchmark` added as a top-level key, applied in `before_instantiate_classes` — mirroring the existing `matmul_precision` mechanism exactly rather than inventing a parallel one. Free at the static 96×96 training crop shapes.
- `ThroughputMonitor` / `DeviceStatsMonitor` as commented-out callback examples.

### Measured: no speedup — and that is the correct result

A synthetic forward+backward benchmark at both templates' real training shapes on the dev GPU (RTX 5060 Laptop, sm_120) showed **no measurable win** from `matmul_precision=high` + `cudnn.benchmark=True` — smaller than run-to-run variance. bf16-mixed autocast was, if anything, marginally slower.

This is expected, not a failed measurement: `torch.set_float32_matmul_precision` governs the **matmul/linear** path (`torch.backends.cuda.matmul.allow_tf32`, observed `False`), while convolutions go through cuDNN, whose `torch.backends.cudnn.allow_tf32` **already defaults to `True`**. SRCNN and SRResNet are pure conv stacks with essentially no matmul, so there was never a TF32 win here to capture.

The levers ship as conventional, harmless defaults. **The template comments describe them as mechanism, not as a speedup** — nothing in this PR claims a win the architecture doesn't have.

### `channels_last`: deliberately not wired

It needs both model and input tensors converted (a code change, not a YAML toggle), its benefit concentrates under AMP which is off by default here, and there is no measured win to extrapolate from. Half-wiring it would ship an unverifiable example.

### `torch.compile`: out of scope

Deferred pending measurement. For the record, the data path — not the GPU — was the real bottleneck: see #36, where decode was **98.94%** of per-sample cost.

### Test note

`test_matmul_precision_defaults_to_none_in_process` asserted the template resolves to `None` when unset, which this PR makes false by design. It was inverted rather than deleted; the distinct "hook no-ops when unset" semantic remains covered by the untouched `test_before_instantiate_classes_skips_when_unset`. Symmetric tests added for `cudnn_benchmark`.

**Tests:** 251 → **255**. Reviewed independently: spec PASS, quality APPROVED, no Critical/Important findings.

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -24,6 +24,11 @@ on Ampere+ GPUs to enable TF32 matmul kernels.
 :func:`sisr.export.to_onnx` — see that module for what gets exported and its
 SRCNN limitation. It requires the optional ``export`` extra
 (``pip install '.[export]'``).
+
+Top-level ``cudnn_benchmark:`` (``true`` / ``false``) sets
+``torch.backends.cudnn.benchmark`` once at startup. Free win whenever input
+shapes are static across steps, as they are for this project's fixed training
+crops (cuDNN autotunes a kernel per shape on first sight and reuses it).
 """
 
 import sys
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #30, opened 2026-07-31
CI, tooling & test hygiene (P5.8, P4.12, P4.13, P5.10, P5.11, INIT.6)

YeapJieShen:
Six independent hygiene items. **No `sisr/` source changes.**

| Item | Change |
|---|---|
| **P4.12** | Remove `-n 2 --dist=loadscope` from `addopts` so a bare local `pytest` runs serial. The measurement comment block is **relocated verbatim** into `test.yml`, not deleted. |
| **P4.12 (CI)** | `test.yml` ran `pytest -n 8` — the count our own measurement recorded as *slower than serial* (`-n 0` 44.0s, `-n 2` 38.3s, `-n 8` 47.1s). Runners have fewer cores than that machine, so aligned to `-n 2 --dist=loadscope`. |
| **P5.8** | `UP038` was deprecated in ruff 0.12.0 and **removed** in 0.16.0 (our pinned version). Listing it under `ignore` suppresses nothing and makes ruff warn on every run. Replaced with a comment on `select`. |
| **P5.10** | Delete two 0-byte `tests/**/__init__.py` files — unnecessary under `--import-mode=importlib` and contradicting the documented convention. |
| **P5.11** | `test_cli_help_lists_subcommands` spawned the `sisr` binary, cold-importing torch+lightning (~10s of a ~38s suite) to print `--help`. Moved to `build.yml`, which already installs and smoke-imports the wheel. |
| **P4.13** | Nothing exercised pip's *editable* path after `test.yml` moved to uv, yet `CONTRIBUTING.md` still gives contributors `pip install -e ".[dev]"`. Adds one cheap `editable-install` leg, wired into the `build` gate's `needs:`. |
| **INIT.6** | `.github/dependabot.yml`, github-actions only, weekly, grouped. |

**Test count:** 217 → **216** (P5.11 relocates one test out of the suite). Full suite green; `ruff check` and `ruff format --check` both clean and warning-free.

Gate job names `test` and `build` are unchanged — they are required-check contexts.

Part of the 2026-07-31 triage delivery arc.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #38, opened 2026-07-31
Add opt-in bare-model ONNX export with CI parity check (INIT.7)

YeapJieShen:
> **Stacked on #30** (`chore/ci-tooling-hygiene`) — merge that first.

### Bare-model export, deliberately

Exports `module.model` only — not model+processor. Consumers have `sisr` importable and call `processor.extract`/`reconstruct` themselves, so the ONNX graph matches `SRLightning.forward` exactly and **makes no claim it cannot keep**.

This is cheap because `RGBProcessor.extract`/`reconstruct` are literal identity functions — for SRResNet the bare graph *is* already end-to-end. The gap exists only for SRCNN + `YChannelProcessor`, whose `reconstruct(sr, lr_rgb)` needs the LR RGB back for chroma: a two-input graph the bare export omits by design. **Documented as a stated limitation** — an SRCNN ONNX consumer outside Python must reimplement Y extraction and chroma upsampling.

### Measured parity

Against onnxruntime **CPU** (GitHub runners have no GPU), exported at one spatial size and inferred at a **different** one, so `dynamic_axes` is genuinely exercised rather than assumed:

| Model | Export → inference | Max abs diff |
|---|---|---|
| SRCNN | 48×48 → 30×70 | **6.7e-08** |
| SRResNet | 16×16 → 20×28 | **1.5e-07** |

Both at float32 machine epsilon.

### Both surfaces

`sisr.export.to_onnx(...)` is the real implementation; `sisr export` is a thin CLI wrapper.

`export` isn't a stock Lightning subcommand, so it needs a home: `LightningCLI` derives each subcommand's parser from a method of that name on `trainer_class`. `_ExportTrainer` exists solely to provide it, and changes nothing for `fit`/`validate`/`test`/`predict`. **No private Lightning APIs are called.**

Because `LightningCLI` resolves a parser for *every* registered subcommand at construction, a `trainer_class` without `export` used to fail as a bare `AttributeError` from inside Lightning. It now raises a `TypeError` naming the class and the fix (`399bfd0`).

**Durability:** that same unconditional resolution means essentially every existing `test_cli.py` test already exercises the export wiring — a Lightning bump breaking the mechanism would fail most of the CLI suite, not slip through silently.

### Optional extra, no import weight

`onnx`/`onnxruntime` live behind a `[export]` extra and are imported lazily. `import sisr`, `import sisr.cli`, and `import sisr.export` all succeed with them absent; attempting an export without them raises an `ImportError` naming the extra. Import weight is a first-class constraint in this repo — spawned processes pay a ~10s torch+lightning cold import.

Uses `torch.onnx.export(..., dynamo=False)` — the legacy exporter, since torch's new default needs `onnxscript`, outside this extra's scope.

### TensorRT: declined

NVIDIA-only, version-brittle, and CI has no GPU, so it would carry zero coverage and rot silently. The README instead documents the ONNX→TRT path via `trtexec`, with `--minShapes`/`--maxShapes` for the dynamic axes — no dependency, nothing to rot.

### CI

New `onnx-export` job feeding the unchanged `test` gate's `needs:`. Gate job names `test`/`build` are untouched (required-check contexts).

**Tests:** 216 → **226**. Docs are a separate commit (`00aea2a`). Reviewed independently: spec PASS, quality APPROVED, two Minor findings fixed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,12 +86,36 @@ jobs:
           path: coverage.xml
           retention-days: 14
 
+  # Numeric-parity check for the opt-in ONNX export (INIT.7): torch eager vs.
+  # onnxruntime CPU (GitHub runners have no GPU), both architectures. Single
+  # OS/Python leg — a correctness check on one code path, not a compatibility
+  # matrix. Installs the `export` extra explicitly since test-matrix's `.[dev]`
+  # deliberately excludes it (onnx/onnxruntime stay opt-in for contributors).
+  onnx-export:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          cache-suffix: onnx-export
+      - name: Install
+        run: uv pip install --system -e ".[dev,export]"
+      - name: ONNX export parity tests
+        run: pytest tests/test_export.py tests/test_cli.py -k export -v
+
   # Stable required-check context: green only if every matrix leg passed
   # (legs report as "test-matrix (os, ver)", which are not fixed contexts).
   test:
-    needs: test-matrix
+    needs: [test-matrix, onnx-export]
     if: always()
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #45, opened 2026-08-02
Add SRProcessor.extract_target for asymmetric model IO ranges

YeapJieShen:
**PR 1 of a queued stack.** Merge in order — later PRs base off this one.

## What

`SRProcessor.extract` served double duty: it produced both the model's *input* and the *loss target*. That silently assumes a model's input space equals its output space — true for every pure colorspace adapter, which is why nothing had exposed it, and false for the SRResNet recipe in Ledig et al. §3.2:

> We scaled the range of the LR input images to [0, 1] and for the HR images to [-1, 1]. The MSE loss was thus calculated on images of intensity range [-1, 1].

This adds a concrete `extract_target(hr_rgb)` defaulting to `self.extract(hr_rgb)`, and rewires `SRLightning._step` to use it. Every existing processor is correct unmodified — additive, no back-compat shim.

`RGBSignedOutputProcessor` is the only override: identity in, `2*hr - 1` as the target, `(out + 1)/2` back out.

## Why not the alternatives

- **A symmetric `[-1,1]` processor** (scaling LR too) needs no code change, but puts the LR input in a range the paper doesn't use — not the thing we set out to reproduce.
- **A range argument on `SRResNet.__init__`** can't work: the model cannot transform the *target*.
- **Leaving it**, on the argument that Adam cancels the 4× loss scaling. Half right — the scaling genuinely does cancel, so `loss/train` reading 4× higher is cosmetic. What doesn't cancel is the target *distribution*: `[-1,1]` is zero-centred, matching what a freshly initialised conv stack emits, where a `[0,1]` target makes the tail conv learn a +0.5 mean shift first.

## Invariant

`extract_target` must be the exact inverse of `reconstruct`. If they drift, the loss and the metrics score different quantities and **nothing fails loudly**. Two tests pin it: a round-trip assertion, and `test_step_loss_uses_extract_target_not_extract`, which proves the loss is 4× the `[0,1]` value rather than equal to it.

## Verification

- `333 passed` (324 → 333; 9 new). Run CPU-only to avoid contending with a live training run.
- `ruff check` clean, `ruff format --check` clean.
- End-to-end: 2 training steps through the real `fit` path with the new processor; `loss/train` opens at ~0.42 where a `[0,1]` run opens near 0.1.

## Note

The tracked SRResNet template still pairs with `RGBProcessor`, so shipped defaults are unchanged by this PR. Switching it is deliberately separate — it renumbers any recorded baseline.

Docs are a separate commit per project convention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,8 +151,13 @@ Consumers are expected to have `sisr` importable and call `processor.extract` /
 computes, rather than silently baking in Python-side pre/post-processing an ONNX runtime
 can't see.
 
-- **SRResNet is unaffected by this**: `RGBProcessor.extract` / `reconstruct` are identity
-  functions, so the exported graph already is the complete LR-RGB → SR-RGB pipeline.
+- **SRResNet with `RGBProcessor` is unaffected by this**: `extract` / `reconstruct` are
+  identity functions, so the exported graph already is the complete LR-RGB → SR-RGB
+  pipeline.
+- **SRResNet with `RGBSignedOutputProcessor` needs one step**: that processor trains the
+  model to emit `[-1, 1]` (the SRGAN paper's HR range), so the graph's output must be
+  rescaled by `(out + 1) / 2` to land back in `[0, 1]`. The input side needs nothing —
+  the model consumes `[0, 1]` either way.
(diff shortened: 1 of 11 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #46, opened 2026-08-02
Trim verbose comments to non-obvious why only

YeapJieShen:
**PR 2 of a stack — bases off #45, not `main`.** Merge #45 first.

## What

Comments and docstrings drifted verbose: multi-paragraph inline essays, docstring prose restating what the next line plainly does, and the `name (type): The <name> to use.` pattern that just re-spells the parameter. This trims to the project's standing rule — **comments explain non-obvious *why* only**.

**15 files, +252 / −510.**

## What was deliberately kept

Terseness was not allowed to cost the hard-won reasons. Explicitly preserved: `sisr/cache.py`'s torch-free constraint, the LMDB `.copy()` use-after-free note in both dataset modules, the deliberately-uncached random crop, `sisr/export.py`'s `dynamo=False` and its scoped warning filters, `SRCheckpoint`'s `/`-sanitisation, and the `validate_against`-not-`__post_init__` reasoning.

`sisr/training/config.py` and the two model `config.py` files were left essentially untouched — their docstrings are dense but carry paper-specific values, not signature restatement.

## Public API kept its `Args:` blocks

An initial pass over-corrected and stripped `Args:` blocks wholesale; that was walked back. The rule applied: **public API keeps a complete `Args:` block** — every parameter documented, wording compressed (dropping the redundant `(type)` restatement the annotation already carries). Private `_`-prefixed helpers may collapse to one line or nothing. Never a partial block.

Worth noting: **ruff cannot enforce this.** `select` has no `D` rules, and even with them, D417 fires only on *partially* documented functions — a fully deleted `Args:` block passes. Verified empirically. Adding scoped `D` rules is queued separately, with that limitation recorded.

## Verification

Beyond the suite, the no-behaviour-change claim is **proven, not asserted** — comments never reach the AST, so every changed `.py` file was parsed at both revisions, docstring nodes stripped, and `ast.dump()` compared:

```
PASS — all 13 changed .py files are AST-identical
```

A second check confirmed the `Args:` walk-back landed: of 70 public callables with parameters, none lost a block it previously had.

- `332 passed, 1 skipped` (the skip is `test_imresize.py`'s byte-equality check — `data/reference` is gitignored and absent from the worktree; it runs in a normal checkout)
- `ruff check` clean, `ruff format --check` clean

## One removal worth flagging

The `(confirmed here: RTX 5060 Laptop, sm_120 Blackwell)` provenance was cut from the `matmul_precision` comment in both templates. That is correct but incidental — the provenance was vouching for a speedup that INIT.12 measured as a **no-op** on these conv stacks. Fixing the surrounding claim is queued as a separate docs change.

### DIFF
--- a/sisr/cache.py
+++ b/sisr/cache.py
@@ -32,8 +32,8 @@ class LMDBCacheBuildContext:
     Provides helpers for writing data into the LMDB being built.
 
     Args:
-        env (lmdb.Environment): The open LMDB environment (writable).
-        use_tqdm (bool): Whether progress bars should be displayed.
+        env: The open, writable LMDB environment.
+        use_tqdm: Whether to display progress bars.
     """
 
     def __init__(self, env: lmdb.Environment, use_tqdm: bool = False):
@@ -44,7 +44,7 @@ class LMDBCacheBuildContext:
         """Writes a batch of key-value pairs in a single transaction.
 
         Args:
-            pairs (Sequence[tuple[str, bytes]]): Sequence of ``(key, value)`` tuples to write.
+            pairs: Sequence of ``(key, value)`` tuples to write.
         """
         txn = self.env.begin(write=True)
         for key, value in pairs:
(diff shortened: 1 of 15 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #47, opened 2026-08-02
Correct SRCNN paper init std to 0.001 (P1.1)

YeapJieShen:
**PR 3 of a stack — bases off #46.** Merge #45 → #46 → this.

## The bug

SRCNN's paper-faithful weight init used a standard deviation **10× too large**, under a docstring asserting it matched the paper. Dong et al. (arXiv 1501.00092), §Training:

> Filter weights of each layer are initialized by drawing randomly from a Gaussian distribution with zero mean and standard deviation **0.001** (and 0 for biases).

The code used `0.01`. That is 10× in std, **100× in variance**. `SRCNNTrainingConfig.init_strategy` defaults to `'paper'`, so it fired on every default SRCNN run — with fan-in 81 on the 9×9 first layer, the initial pre-activation std moves ~0.009 → ~0.09, into a plain SGD+momentum optimiser with no adaptive rescaling to absorb it.

**The docstring is why this survived.** It claimed fidelity, so nobody re-derived it. That is the real defect class here, not the digit.

## Where the constant lives

`0.001` is an *SRCNN paper value*, but `init_std` sits on the generic `SRTrainingConfig` base shared by every architecture. Putting a paper constant there would be the same mistake one level down.

So: an explicit `init_std: float = 0.001` **override on `SRCNNTrainingConfig`**, plus `SRCNN.reset_parameters`'s own default (SRCNN-specific code). The shared base keeps `0.01` — it has no paper to match — and its docstring, which claimed these were "matching the SRCNN paper", now says plainly that the base is not paper-derived and that architectures override it with their real value.

## SRResNet: confirmed unaffected, by tracing not assumption

`SRLightning.__init__` calls `reset_parameters(mean=…, std=…)` only when `init_strategy == 'paper'`; `SRResNetTrainingConfig` defaults to `'default'`, so it never fires. And `sisr/models/srresnet/model.py` has no `reset_parameters` override, so it inherits `SRModel`'s `**kwargs`-absorbing no-op — the values would be ignored even if flipped. Doubly inert.

## Tests

- **Updated:** `test_srcnn_training_config_init_defaults` asserted the old `0.01`. Other `0.01` occurrences were checked and left — they test the *base* default or an explicitly-passed value, not the SRCNN paper default.
- **Added:** `test_reset_parameters_default_std_matches_paper_not_tenfold_larger` — pools weights across every `Conv2d` (~20k elements) and asserts the realised std falls in `(0.0008, 0.0012)`. With that many pooled samples the sampling-noise band is ~±0.5%, so the ±20% tolerance is ~40× the noise floor and won't flake, while the old `0.01` lands ~8× above the upper bound and fails decisively.

Both halves are covered: the config override *and* the model's own default.

## Verification

- `333 passed, 1 skipped` (baseline 332 + 1 new; skip is `test_imresize.py`'s byte-equality check, `data/reference` absent from worktrees)
- `ruff check` clean, `ruff format --check` clean
- Run with `CUDA_VISIBLE_DEVICES=""` — a training run owns the GPU

Doc-only changes are a separate commit per project convention.

### DIFF
--- a/sisr/models/srcnn/config.py
+++ b/sisr/models/srcnn/config.py
@@ -26,7 +26,7 @@ class SRCNNTrainingConfig(SRTrainingConfig):
     uses a per-layer learning rate of ``1e-4`` for the feature-extraction
     and non-linear-mapping layers and ``1e-5`` for the reconstruction layer
     — i.e. the last layer learns 10× slower. Weight initialization follows
-    the paper's Gaussian schedule (``N(0, 0.01)`` with zero biases); set
+    the paper's Gaussian schedule (``N(0, 0.001)`` with zero biases); set
     ``init_strategy='default'`` to fall back to PyTorch's built-in init.
     Override any field in YAML to deviate.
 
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #48, opened 2026-08-02
Report Y-channel PSNR by default and ship the paper's range (P2.9, P2.10)

YeapJieShen:
Two changes that make the shipped defaults match the papers.

## P2.9 — neither model emitted a paper-comparable PSNR

Both eval configs defaulted to `psnr_channels=['RGB','YCbCr']` with `separate_psnr=False`, so the only luma-ish key was `psnr/val/YCbCr` — PSNR over the **three-channel** tensor, `MSE = (MSE_Y + MSE_Cb + MSE_Cr)/3`. Chroma planes are far smoother than luma, so that aggregate reads optimistically high and is not the metric either paper reports. Both quote **Y-channel only**.

**Fix: `psnr_channels` now accepts a bare channel entry**, and both defaults become `['RGB', 'Y']` → exactly `psnr/val/RGB` and `psnr/val/Y`.

The alternative — flipping `separate_psnr` to `True` — also works but drags in R/G/B/Cb/Cr as collateral, and conflates "give me per-channel breakdowns" with "give me the paper's metric".

The implementation needed **no new logic**. `_PSNR_CHANNEL_NAMES` is the single source of truth for both the validation allowlist and key expansion, so each bare channel simply maps to `()`:

```python
"RGB": ("R", "G", "B"), "YCbCr": ("Y", "Cb", "Cr"),
"R": (), "G": (), "B": (), "Y": (), "Cb": (), "Cr": (),
```

`psnr_keys`' existing `extend(...)`-then-`append(cs)` then yields exactly one key for a bare channel. `_build_psnr_tensors` already precomputed `Y`/`Cb`/`Cr` whenever any of `{YCbCr, Y, Cb, Cr}` was requested, so it was untouched too.

All three `psnr_keys` consumers were checked, not assumed: `SRLightning` (val logging + HParams), `BenchmarkImageLogger`, and `SRCheckpoint.setup`'s monitor validation.

## P2.10 — the paper's range was not what shipped

`RGBSignedOutputProcessor` existed but the tracked SRResNet template still paired with `RGBProcessor`, so an out-of-the-box `sisr fit` trained entirely in `[0,1]`. Template now points at the signed processor, with a two-line note that `RGBProcessor` reproduces pre-2026-08-02 runs.

`config.srcnn.template.yaml` is untouched — SRCNN trains Y in `[0,1]` via `YChannelProcessor` and is already correct.

Verified by `--print_config`: processor resolves to `RGBSignedOutputProcessor`, eval config to `psnr_channels: [RGB, Y]`, `crop_border: 4`.

## Note: this renumbers recorded SRResNet baselines

Expected and one-time. Worth knowing what the numbers should look like — a run on the previous defaults plateaued at Set5 30.56 raw Y-PSNR, which under P2.8's still-open full-range convention is `+1.3225 dB` from the literature's studio-range figure, i.e. ~31.89 vs the paper's 32.05. The new `psnr/val/Y` key reports that same raw value; **P2.8 is what will make it directly comparable.**

## Tests

`340 passed, 1 skipped` (334 baseline + 6). Existing tests pinning the old defaults were updated — the parametrized `psnr_keys` test gained bare-channel cases rather than being duplicated. New: one confirming a bare `'Y'` yields exactly one key (no decomposition, unlike `YCbCr`), and one confirming the widened allowlist did **not** loosen validation — `'HSV'` still raises.

`ruff check` and `ruff format --check` clean.

### DIFF
--- a/sisr/models/srcnn/config.py
+++ b/sisr/models/srcnn/config.py
@@ -80,16 +80,18 @@ class SRCNNTrainingConfig(SRTrainingConfig):
 class SRCNNEvalConfig(SREvalConfig):
     """SRCNN-paper-faithful eval defaults.
 
-    Reports PSNR on both RGB and the YCbCr triplet (the literature
-    usually quotes Y-channel PSNR for SRCNN), and excludes the outer
-    ``scale=3`` pixels per the standard SR-evaluation convention.
+    Reports PSNR on RGB and Y — the paper's own metric (Dong et al. quote
+    Y-channel only; the full YCbCr aggregate reads optimistically high
+    since chroma planes are far smoother than luma) — and excludes the
+    outer ``scale=3`` pixels per the standard SR-evaluation convention.
 
     Args:
         crop_border: Overrides the base default to ``3`` (outer pixels
             excluded before PSNR / SSIM at the standard ``x3`` scale).
-        psnr_channels: Overrides the base default to
-            ``['RGB', 'YCbCr']`` (the literature reports both).
+        psnr_channels: Overrides the base default to ``['RGB', 'Y']`` —
+            ``'Y'`` is the paper's own metric; ``'RGB'`` is a supplementary
+            aggregate.
     """
 
     crop_border: int = 3
-    psnr_channels: list[str] = field(default_factory=lambda: ["RGB", "YCbCr"])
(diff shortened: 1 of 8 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #49, opened 2026-08-02
Correct paper claims and shipped-doc drift (P5.12, P5.13)

YeapJieShen:
Docs-only. Every item is a factual claim in a tracked file that was wrong.

## 1. A fabricated paper claim

`SRResNetTrainingConfig`'s docstring asserted *"Weight initialization in the paper is implicit (Kaiming-style for PReLU activations)."* **Verified against the full text of Ledig et al.: the paper contains no discussion of weight initialization anywhere.** Now states that plainly and marks the Kaiming idea as a project choice reserved for future work, not the paper's.

This sat in a file whose entire job is recording what the paper specifies — the same failure mode as P1.1, where a docstring claiming `0.01` "matched the SRCNN paper" kept anyone from re-deriving a 10x-wrong value.

## 2. `CONTRIBUTING.md` undercounted required checks

Said **two** (`test`, `build`). The `main-protection` ruleset has required **three** since 2026-07-31 — `lint` is its own top-level job, confirmed against `.github/workflows/test.yml` rather than assumed.

## 3. Incomplete processor list, in two files

`README.md` and `CONTRIBUTING.md` both listed three processors; `RGBSignedOutputProcessor` was missing. README's own ONNX section already discussed it, so the file contradicted itself. Authoritative set taken from `__all__`.

## 4. README overstated paper comparability

Claimed byte-exact `imresize` *"is what makes PSNR/SSIM numbers comparable to published papers."* Byte-exact **inputs** do not imply comparable **reported numbers** while P2.8 is open. Softened to a claim about the LR inputs, with the metric-convention caveat named as what it is — a known, exact `20*log10(255/219) ~= 1.3225 dB` offset, not an unknown.

## 5. Both templates oversold `matmul_precision`

Described as *"a standard free lever"*. It was **measured as a no-op** on these conv stacks: `set_float32_matmul_precision` governs matmul, while conv goes through cuDNN whose `allow_tf32` already defaults `True`. The key is kept (harmless); the claim of benefit is gone.

## 6. The processor now selects range, not just colorspace

`SRResNetTrainingConfig` named only `RGBProcessor` as the paper's pairing. That was complete when colorspace was a processor's whole job — it now reads as endorsing the wrong one, since `RGBSignedOutputProcessor` carries the paper's `[-1,1]` output range.

## Verification

`334 passed, 1 skipped` — **unchanged**, which is the point: any movement would mean a docs-only PR touched behaviour. `ruff check` / `ruff format --check` clean. Both templates re-verified through `--print_config`.

Deliberately not changed: `sisr/cli.py`'s `matmul_precision` docstring describes only the mechanism and never claims a benefit for this repo, so it isn't false.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,10 +86,11 @@ parallel data loading; use it only for debugging, not for real training runs.
 3. Run `pytest` locally and make sure it is green.
 4. Open a pull request against `main`.
 
-All PRs must pass two GitHub Actions checks before they can merge:
+All PRs must pass three GitHub Actions checks before they can merge:
 
 - **test** — installs `.[dev]` and runs `pytest` with coverage (uploaded to Codecov).
 - **build** — verifies `pip install .` and that `sisr` / `sisr.cli.main` import cleanly.
+- **lint** — `ruff check` and `ruff format --check`.
 
 `main` uses linear history (rebase/squash, no merge commits) and requires PRs.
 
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #50, opened 2026-08-02
Compute Y-channel metrics in BT.601 studio range and report Y SSIM (P2.8, P3.8)

YeapJieShen:
The metric half of paper comparability. The arc made the *inputs* byte-exact; nothing had checked the convention the *outputs* are computed in.

## P2.8 — the metric used the wrong BT.601 variant

`sisr/colorspace.py` is BT.601 **full-range** ("JPEG"). Every published SR figure — MATLAB `rgb2ycbcr`, BasicSR, the `daala` tools Ledig et al. used — is **studio/limited range**. The error term scales by a constant, so a full-range Y-PSNR sits exactly `20·log10(255/219) = 1.3219 dB` below the literature's.

**Not a flag on the existing function — a separate `rgb_to_ycbcr_studio`.** `rgb_to_ycbcr` is also `YChannelProcessor`'s **training** colorspace, so a default-off keyword would leave a live footgun: a future caller flips it, or someone "unifies" the default, and a trained model's inputs silently renumber. A distinctly named one-way function makes that impossible by construction. The module docstring now says which range is training-side and which is metric-side, and not to merge them.

Routed through the single shared tensor-map builder, so `SRLightning` and `BenchmarkImageLogger` both pick it up — no second code path.

### Correction to the brief I wrote

I specified that the `20·log10(255/219)` identity should hold across the whole Y/Cb/Cr family. **That is wrong and the implementer was right to refuse it.** BT.601 studio range uses **219/255 for luma but 224/255 for chroma**. Chroma's true delta is `20·log10(255/224) ≈ 1.1258 dB`, and the 3-channel aggregate isn't a constant at all. The correct per-channel scales are implemented, the identity test is scoped to Y, and a companion test locks in that chroma deliberately differs.

Verified against MATLAB's published coefficients (`Y = 16 + 65.481R + 128.553G + 24.966B`, etc.):

```
Y  max|diff| = 3.3e-16   <- exact
Cb max|diff| = 2.3e-04
Cr max|diff| = 2.7e-04

Y  studio-full = +1.32192 dB   expected 20*log10(255/219) = +1.32192  OK
Cb studio-full = +1.12584 dB   expected 20*log10(255/224) = +1.12584  OK
```

**Y — the channel the papers actually quote — matches MATLAB exactly.** The chroma residual is inherited from `sisr/colorspace.py`'s 3-decimal BT.601 chroma coefficients (`-0.169` vs MATLAB's `-0.168737`), a pre-existing property this PR doesn't introduce and doesn't worsen. Filed separately.

### Also: the quoted constant was wrong

I had been writing **1.3225 dB**; it is **1.3219**. The computed assertions were always right — only the prose restating them was wrong, including in the README this repo shipped last hour. Corrected here. Same failure mode as P1.1: a stated constant nobody re-derives because it looks checked.

## P3.8 — Y-channel SSIM

SSIM was hardcoded to RGB in two places. The papers quote **y-channel** SSIM, and unlike PSNR there's no constant to apply afterwards — there was simply no comparable number available.

Adds `SREvalConfig.ssim_channels`, mirroring `psnr_channels` and reusing the same channel machinery and tensor map rather than growing a parallel system. Defaults to `['RGB', 'Y']` **on the base**, since SSIM has no post-hoc correction and there's no reason to gate it behind an architecture subclass.

**Key shape changes: `ssim/val` → `ssim/val/{key}`.** That breaks three things, all updated: `SRCheckpoint`'s valid-metric set, both templates' `monitor_metric: ssim/val`, and `BenchmarkImageLogger`'s per-sample scalar (now a per-key dict, matching how PSNR already worked). Every `ssim/` occurrence was grepped before and after.

## Verification

- `358 passed, 1 skipped` (340 baseline). Each commit independently green — verified by isolating them, not assumed.
- `ruff check` / `ruff format --check` clean.
- Both templates re-resolved: `monitor_metric: ssim/val/RGB`, `ssim_channels` present.

## Note

This changes what `psnr/val/Y` **means** — same key, different convention. Runs before and after this PR are not comparable on that key despite the matching name. An existing run plateaued at Set5 30.56 raw; after this PR the equivalent should read ≈ 31.88 natively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ via `resize_backend`:
   Byte-exact inputs alone don't make **reported** PSNR/SSIM comparable, though:
   Y-channel figures here use BT.601 full-range Y (`sisr/colorspace.py`), while the
   literature's published Y-channel numbers use MATLAB `rgb2ycbcr`'s studio-range
-  convention — a known, exact `20·log10(255/219) ≈ 1.3225 dB` offset, not an unknown one.
+  convention — a known, exact `20·log10(255/219) ≈ 1.3219 dB` offset, not an unknown one.
 - **`'cv2'` (opt-in).** Plain `cv2.INTER_CUBIC`, no antialiasing of its own. SRCNN's
   `blur_sigma` (a pre-resize Gaussian blur) only has an effect on this path — passing
   `blur_sigma` together with `resize_backend='matlab'` raises `ValueError` at
--- a/sisr/colorspace.py
+++ b/sisr/colorspace.py
@@ -1,9 +1,21 @@
-"""BT.601 full-range RGB <-> YCbCr conversion.
+"""BT.601 RGB <-> YCbCr conversion, full-range and studio-range.
 
-BT.601 full-range YCbCr is the project's working chroma space. These pure
-tensor functions live on their own so the
-:class:`~sisr.processors.SRProcessor` subclasses and any external callers can
-convert colorspaces without depending on Lightning or model code.
+Two distinct ranges live here, deliberately kept apart:
+
+* **Full-range** (:func:`rgb_to_ycbcr` / :func:`ycbcr_to_rgb`) is the
+  project's *training* chroma space — what :class:`~sisr.processors.SRProcessor`
+  subclasses (``YChannelProcessor``, ``YCbCrProcessor``) feed the model.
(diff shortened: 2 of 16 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #51, opened 2026-08-02
Clamp SR output to [0, 1] before scoring (P4.14)

YeapJieShen:
PSNR/SSIM scored the raw network output, which overshoots `[0, 1]`. Published SR numbers are computed on 8-bit images, i.e. implicitly clamped. Overshoot is pure penalty — clamping can only help — so our figures sat below what the same weights would score under the standard protocol.

**One line of behaviour change**, in `SRLightning._forward_lr`, immediately after `processor.reconstruct(...)`.

## Why there, specifically

`_forward_lr` is the **only** place `reconstruct()` is ever called. Every consumer reads through it — `_forward_sr` (which feeds `_step`, so both `training_step` and `validation_step`) and `predict_rgb` (which feeds `BenchmarkImageLogger`), plus `predict_step` directly. One line covers all of them with no parallel edit.

Placing it one level up in `_forward_sr` would have been wrong, and an **existing invariant test caught it**: `test_forward_lr_matches_forward_sr_sr_component` asserts the two paths are bit-identical. That equality is not vacuous — the seeded fixture already overshoots to `-0.127`, so clamping in only one path would have reintroduced exactly the path-divergence the codebase deliberately closed.

## The four constraints, each verified rather than assumed

1. **The loss is untouched.** `sr_model_out` is returned separately and never clamped — clamping it would kill gradients on saturated pixels. A test biases a tail conv to force genuine overshoot, then asserts the loss equals `mse_loss(sr_model_out, hr_cropped)` exactly *and* differs from `mse_loss(sr_rgb, hr_cropped)`, which is what a leaked clamp would produce.
2. **`SRResNet.forward`'s `clamp_output`/`clamp_minmax` untouched.** Their correct bounds depend on the training processor; using them here would reintroduce that coupling. No model file is in the diff.
3. **No processor touched.** `YCbCrProcessor` already clamps via `ycbcr_to_rgb` while the RGB processors do not — that asymmetry (SRCNN accidentally clamped, SRResNet not) *is* the bug. Fixing it in a processor would make clamping depend on training colourspace. The new clamp is idempotent on the already-clamped path.
4. **Both metric consumers covered**, each with its own test — `validation_step` and `BenchmarkImageLogger`'s seam.

RED→GREEN was confirmed by commenting the clamp out: all four new tests fail, one showing PSNR at `-13.07` unclamped versus `4.997` clamped.

## Measured effect

- Forced overshoot (output range `[4.81, 5.34]`): **−13.13 dB → +5.00 dB**. The clamp turns a nonsensical negative score into a coherent one.
- Realistic untrained SRResNet (mild overshoot to `1.027`): **+0.0001 dB**. Small because random-init weights barely saturate; the 0.05–0.2 dB figure applies to trained models.

## Note

This changes reported numbers — upward, by removing a penalty. Combined with #48 and #50, `psnr/val/Y` has now changed meaning three times in this arc, so a run from before is not comparable to one after despite the identical key name.

`362 passed, 1 skipped` (358 baseline, +4 new, **0 existing tests modified**). `ruff check` / `ruff format --check` clean.

### DIFF
--- a/sisr/training/lightning_module.py
+++ b/sisr/training/lightning_module.py
@@ -225,11 +225,20 @@ class SRLightning(lightning.LightningModule):
 
         Returns:
             ``(sr_model_out, sr_rgb)`` — the raw model output in the model IO
-            colorspace, and the reconstructed SR RGB.
+            colorspace, and the reconstructed SR RGB clamped to ``[0, 1]``.
         """
         model_input = self.processor.extract(lr_img)
         sr_model_out = self.model(model_input)
         sr_rgb = self.processor.reconstruct(sr_model_out, lr_img)
+        # Clamp display-space output only, here, once — every reconstruct()
+        # consumer (_forward_sr's callers, predict_step) reads through this
+        # one line, so PSNR/SSIM never diverge from what an 8-bit image would
+        # score (P4.14). sr_model_out (the loss target) is untouched: clamping
+        # it would kill gradients on saturated pixels. Idempotent where
+        # reconstruct() already clamps (YCbCrProcessor/YChannelProcessor's
+        # ycbcr_to_rgb) — don't move this into forward() or a processor, or
+        # the bound would depend on training colorspace again.
+        sr_rgb = sr_rgb.clamp(0.0, 1.0)
         return sr_model_out, sr_rgb
 
     def _forward_sr(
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #52, opened 2026-08-02
Enforce Google-style docstrings via ruff pydocstyle (P4.15)

YeapJieShen:
"Google-style docstrings on public API" is a stated project non-negotiable that was enforced by **review alone**. This makes it a gate.

Not theoretical: a comment sweep earlier in this arc stripped `Args:` blocks from public callables and would have passed CI untouched.

## Config

```toml
select = ["E", "F", "I", "UP", "B", "D"]
ignore = ["D105", "D107"]

[tool.ruff.lint.pydocstyle]
convention = "google"

[tool.ruff.lint.per-file-ignores]
"tests/**" = ["D"]
```

**`D107` (undocumented `__init__`) ignored** — and this is the one worth getting right, because it's the largest single category (19 of 43). Google style documents constructor arguments in the **class** docstring, which this repo already does. Requiring an `__init__` docstring as well would push against the very convention being enforced, for 19 pointless additions.

**`D105`** — dunder semantics are defined by the language, not by us.

**`tests/**` exempt** — 403 violations there, 135 of them undocumented test functions whose *names* are the documentation.

## Violations: measured, then fixed properly

`sisr/` **43** before → **0** after. With the two ignores applied, exactly **19** needed real work: 13 D102, 5 D205, 1 D301.

- **D102** — every one is an override of `SRProcessor`'s abstract interface (`extract`, `reconstruct`, `extract_target`, `model_channels`). Required public contract members, not helpers masquerading as public, so each got a real one-line summary — e.g. *"Map HR from `[0, 1]` to the model's `[-1, 1]` output range."*
- **D205** — these were summary sentences wrapped onto a second physical line. pydocstyle only treats the *first* line as the summary, so each was tightened to one line with the detail moved into the following paragraph — rather than inserting a blank line mid-sentence, which would satisfy the rule while making the docstring worse.
- **D301** — `sisr/cli.py`'s module docstring now uses `r"""`, with identical rendered content.

**No `# noqa` anywhere, and `ignore` was not widened to make numbers disappear.**

## The limitation is documented in the config, deliberately

Recorded inline in `pyproject.toml`, matching how that file already documents its UP038 history:

> **D417 fires only on *partially* documented functions, not undocumented ones.** A docstring with its whole `Args:` section deleted passes with `D` fully enabled — verified with a two-function probe; preview `DOC` rules were silent too. So this catches half-documented functions and formatting drift, **not** the wholesale deletion that motivated adding it. That guardrail stays human review.

Without that written down, the next person assumes the lint covers it. Confirmed live by injecting a partial `Args:` block: `D417 Missing argument description ... 'beta'`.

## Verification

- `ruff check .` clean **with `D` enabled** — the point of the PR.
- `ruff format --check` clean.
- `362 passed, 1 skipped` — unchanged, as docstring edits should be.

Config and fixes are separate commits.

## Reviewer note

The implementer flagged, fairly, that `"tests/**" = ["D"]` is broader than its stated rationale: it exempts D100/D101 and the D200-family too, not just the D103 name-is-the-doc case. Left as-is — the 403 are dominated by D103/D205/D209 and a partial carve-out would be fussier to maintain — but the gap is named rather than silently absorbed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,10 +128,31 @@ line-length = 100
 # is nothing to suppress. Do not add it to `ignore`: ruff warns "rules have been
 # removed and ignoring them has no effect" on every run, which is noise in a required
 # check. The isinstance(x, A | B) style it enforced is now convention, not lint.
-select = ["E", "F", "I", "UP", "B"]
+# D: pydocstyle (P4.15) — makes "Google-style docstrings on public API" (a stated
+# project non-negotiable) enforceable instead of review-only. D105 (undocumented
+# magic method) is ignored because dunder semantics are defined by the language,
+# not by us. D107 (undocumented __init__) is ignored because Google style documents
+# constructor args in the *class* docstring, which this repo already does —
+# requiring an __init__ docstring too would push against the very convention being
+# enforced.
+# Limitation worth knowing: D417 (missing Args entry) fires only on *partially*
+# documented functions, not undocumented ones — a docstring with its whole Args:
+# section deleted passes with D fully enabled (verified empirically with a
+# two-function probe; preview DOC rules were silent on it too). So this catches
+# half-documented functions and formatting drift, not the wholesale Args: deletion
+# that motivated adding it. That guardrail stays human review.
+select = ["E", "F", "I", "UP", "B", "D"]
+ignore = ["D105", "D107"]
+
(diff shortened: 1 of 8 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #53, opened 2026-08-02
Wire BSD100 into default templates and imresize evidence (INIT.19)

YeapJieShen:
## Summary

Ledig et al. Table 2 (SRResNet x4) reports Set5, Set14 **and BSD100**
(27.58 / 0.7620), but this repo only ever benchmarked the first two —
`BSD100_HR` (100 PNGs) was already fetched and sitting unused on disk.

- **`templates/config.srcnn.template.yaml` / `config.srresnet.template.yaml`**:
  add a `BSD100` entry to `test_datasets`, mirroring the existing
  `Set5`/`Set14` entries exactly (same `class_path`, same `init_args`
  shape, `img_dir: data/BSD100_HR`, `scale: *scale`).

  This ships in the shipped default templates rather than a separate
  benchmarking config. BSD100 images are ~0.15 MP each vs DIV2K's
  ~2 MP, so evaluating all 100 of them costs roughly what 7 DIV2K
  validation images cost — the runtime objection that made this a
  judgement call doesn't survive the numbers.

- **`tests/test_imresize.py`**: the MATLAB reference pairs for B100
  are already on disk at `data/reference/B100/{HR,LR_bicubic/X{2,3,4}}`
  (400 files, same EDSR `benchmark.tar` lineage as Set5/Set14), but
  `_reference_cases()` hardcoded `("Set5", "Set14")`. Adding `"B100"`
  widens the byte-equality evidence from **57 to 357 cases**
  (Set5: 15, Set14: 42, B100: 300). Also updated the module docstring's
  extraction instructions to name B100 in both the path listing and
  the example `tar` command, folding in two Windows gotchas:
  `--force-local` (otherwise a `C:` path is parsed as a remote host)
  and `--wildcards` (required for the subtree globs).

## Test plan

- [x] `tests/test_imresize.py -v`: 12 passed, including
      `test_matlab_imresize_matches_real_matlab_reference_data`
      actually running (not skipped) over 357 cases — confirmed via
      `len(_reference_cases())`.
- [x] Full suite: `363 passed` (unchanged count/baseline — this PR adds
      no new test functions, only more parametrized cases inside the
      existing one).
- [x] Both templates resolve through jsonargparse
      (`sisr.cli fit --config ... --print_config`) with `BSD100`
      present in `test_datasets` for both SRCNN (scale 3) and
      SRResNet (scale 4).
- [x] `ruff check .` and `ruff format --check .` both clean.

### DIFF
--- a/templates/config.srcnn.template.yaml
+++ b/templates/config.srcnn.template.yaml
@@ -56,6 +56,11 @@ data:
       init_args:
         img_dir: data/Set14_HR
         scale: *scale
+    BSD100:
+      class_path: sisr.datasets.srcnn.ValidationDataset
+      init_args:
+        img_dir: data/BSD100_HR
+        scale: *scale
   train_dataloader_kwargs:
     batch_size: 64
     num_workers: 16
--- a/templates/config.srresnet.template.yaml
+++ b/templates/config.srresnet.template.yaml
@@ -58,6 +58,11 @@ data:
       init_args:
         img_dir: data/Set14_HR
         scale: *scale
+    BSD100:
+      class_path: sisr.datasets.srresnet.ValidationDataset
+      init_args:
+        img_dir: data/BSD100_HR
+        scale: *scale
   train_dataloader_kwargs:
     batch_size: 16
     num_workers: 16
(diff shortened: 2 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #54, opened 2026-08-02
Use full-precision BT.601 chroma coefficients

YeapJieShen:
## Summary

`sisr/colorspace.py` stored its BT.601 Cb/Cr coefficients truncated to 3 decimal places, missing MATLAB's published `rgb2ycbcr` matrix by up to **~3.1e-4**. This replaces them with full-precision values:

- `_RGB_TO_CB`, `_RGB_TO_CR`: expressed as exact ratios of MATLAB's published studio-range matrix (`65.481/128.553/24.966`, `-37.797/-74.203/112.000`, `112.000/-93.786/-18.214`) divided by the chroma legal-range scale of **224** — kept as literal ratios (not pre-divided decimals) so the MATLAB source constants stay visible in the code and transcription can't silently drift.
- `_YCBCR_TO_G_CB` / `_YCBCR_TO_G_CR` (the inverse's G-channel coefficients): extended from 3 to 6 decimal places (`-0.344` → `-0.344136`, `-0.714` → `-0.714136`), the standard "exact" BT.601 precision.
- `_RGB_TO_Y` and the inverse's R/B coefficients (`_YCBCR_TO_R = 1.402`, `_YCBCR_TO_B = 1.772`) are **unchanged** — they were already exact (`Kr=0.299`, `Kb=0.114` give `2*(1-Kr)=1.402`, `2*(1-Kb)=1.772` exactly).

### Coefficient diff vs MATLAB's matrix

| | old (truncated) | new (full-precision) | max\|diff\| |
|---|---|---|---|
| Y | `(0.299, 0.587, 0.114)` | unchanged | `0.0` (exact) |
| Cb | `(-0.169, -0.331, 0.500)` | `(-0.168737, -0.331263, 0.500)` | `2.634e-4` |
| Cr | `(0.500, -0.419, -0.081)` | `(0.500, -0.418688, -0.081313)` | `3.125e-4` |

### Round-trip tolerance

`tests/test_colorspace.py::test_round_trip_within_coefficient_precision`'s tolerance (whose comment attributed a "~5e-4 error floor" to exactly this truncation) tightened from **5e-4 → 2e-6**. Measured worst case is **~1.31e-6**, a bound confirmed exact by evaluating all 8 vertices of the RGB unit cube (the round trip is affine, so its extrema land on vertices). The residual is no longer float-precision noise but the inherent inconsistency of MATLAB's own 3-decimal-published forward matrix against the analytically-exact `Kr`/`Kg`/`Kb`-derived inverse — tightening further would require abandoning the published MATLAB constants, so 2e-6 is the honest floor. The same two round-trip tests in `tests/processors/test_y_channel.py` and `tests/processors/test_ycbcr.py` (previously `atol=1e-3`, also citing the truncation) were tightened to `2e-6` as well, and one hand-computed expected-value test in `test_ycbcr.py` had its literal `-0.344`/`-0.714` updated to match the module's new constants (it happened to still pass before, by a coefficient-difference cancellation specific to that test's symmetric `cb=-cr` inputs — not a general guarantee).

Added `test_rgb_to_ycbcr_coefficients_match_matlab_matrix`, asserting `_RGB_TO_Y`/`_RGB_TO_CB`/`_RGB_TO_CR` against MATLAB's published matrix to `abs=1e-12`, so the truncation can't silently return.

### Two consequences

1. **This changes SRCNN's training colorspace.** `rgb_to_ycbcr` (full-range) is what `YChannelProcessor`/`YCbCrProcessor` feed the model, so any previously-trained SRCNN checkpoint's Cb/Cr inputs are renumbered by the diffs above. There is no SRCNN checkpoint worth preserving, so this is acceptable — flagging it explicitly rather than glossing over it.
2. **Published metrics are unaffected.** Every paper-comparable figure is Y-channel PSNR/SSIM, and luma was already exact. Verified directly: converting the same random RGB batch through old vs. new coefficients gives a Y-channel max\|diff\| of **exactly 0.0** (bit-identical); only Cb (`2.62e-4`) and Cr (`3.10e-4`) move. Only Cb/Cr PSNR and the non-standard 3-channel YCbCr aggregate are affected.

The two BT.601 variants (`rgb_to_ycbcr` full-range vs. `rgb_to_ycbcr_studio` metric-only) are untouched and not unified — this PR only tightens the shared full-range coefficients both build on.

## Test plan

- [x] `PYTHONPATH="$PWD" python -m pytest -q` → **363 passed, 1 skipped** (the skip is `test_imresize.py`'s reference-data check, expected in a worktree without `data/`)
- [x] `ruff check .` → all checks passed
- [x] `ruff format --check .` → 57 files already formatted

### DIFF
--- a/sisr/colorspace.py
+++ b/sisr/colorspace.py
@@ -28,11 +28,15 @@ import torch
 # [16/255, 235/255] and Cb/Cr to [16/255, 240/255].
 
 _RGB_TO_Y = (0.299, 0.587, 0.114)
-_RGB_TO_CB = (-0.169, -0.331, 0.500)  # output offset +0.5
-_RGB_TO_CR = (0.500, -0.419, -0.081)  # output offset +0.5
+# Ratios of MATLAB's published rgb2ycbcr studio-range matrix (see
+# rgb_to_ycbcr_studio below) over the chroma legal-range scale of 224 —
+# kept as literal ratios rather than pre-divided decimals so the MATLAB
+# source constants stay visible and transcription cannot silently drift.
+_RGB_TO_CB = (-37.797 / 224, -74.203 / 224, 112 / 224)  # output offset +0.5
+_RGB_TO_CR = (112 / 224, -93.786 / 224, -18.214 / 224)  # output offset +0.5
 _YCBCR_TO_R = 1.402  # cr coefficient
-_YCBCR_TO_G_CB = -0.344  # cb coefficient
-_YCBCR_TO_G_CR = -0.714  # cr coefficient
+_YCBCR_TO_G_CB = -0.344136  # cb coefficient
+_YCBCR_TO_G_CR = -0.714136  # cr coefficient
 _YCBCR_TO_B = 1.772  # cb coefficient
 
 # BT.601 studio-range ("limited"/"TV" range) rescale, applied on top of the
--- a/tests/processors/test_y_channel.py
+++ b/tests/processors/test_y_channel.py
@@ -42,9 +42,9 @@ def test_y_channel_reconstruct_same_size_roundtrips_to_lr():
         rgb_to_ycbcr(lr)[:, 1:], size=(16, 16), mode="bicubic", align_corners=False
     )
     assert torch.equal(same, rgb_to_ycbcr(lr)[:, 1:])  # bicubic@same-size == identity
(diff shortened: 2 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #55, opened 2026-08-02
Drop the cv2 degradation backend

YeapJieShen:
## Summary

- Removes the `'cv2'` resize backend entirely; `sisr/imresize.py`'s `resize()` is now MATLAB-only (the `ResizeBackend` alias, `backend` parameter, cv2 branch, and `import cv2` are all gone; `matlab_imresize` and the vendored algorithm are untouched).
- Removes `resize_backend` from both `sisr/datasets/srcnn.py` and `sisr/datasets/srresnet.py`, and removes SRCNN's `blur_sigma`/`_check_blur_sigma()`/`cv2.GaussianBlur` call — blur only ever stood in for cv2's missing antialiasing, so it has no purpose once that backend is gone.
- **Drops the `opencv-python-headless` dependency from `pyproject.toml`.** Verified zero remaining `cv2` call sites under `sisr/` and confirmed every `cv2` reference in `tests/` was testing the cv2 backend itself (including the pass-through oracle), so it wasn't needed as a dev dependency either.
- Updates both templates' comment blocks and the README's degradation-protocol section to describe the single MATLAB path instead of a choice between backends.
- Re-expresses the "LMDB cache checksum is independent of degradation backend" test invariant against a surviving parameter (two datasets differing only in `scale`/`hr_crop_size` now assert equal checksums) instead of letting that coverage disappear.

This is a pre-release repo with a clean-break, no-back-compat-shims policy — the cv2 backend existed only to reproduce caches built before the vendored MATLAB `imresize` landed, which is no longer a live concern.

## Test plan

- [x] `git ls-files | xargs grep -n "cv2\|resize_backend\|blur_sigma\|ResizeBackend"` returns nothing.
- [x] Full suite: `347 passed, 1 skipped` (down from a 363-test baseline: 15 cv2/backend-specific tests removed, and `test_imresize.py`'s byte-equality check skips in this worktree since `data/` isn't present — 363 − 15 = 348 = 347 passed + 1 skipped).
- [x] Both templates resolve: `python -m sisr.cli fit --config templates/config.srcnn.template.yaml --print_config` and the srresnet equivalent both exit 0.
- [x] `grep -rn "import cv2" sisr/` empty — package has no remaining cv2 dependency at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,28 +28,18 @@ watch reconstruction quality improve run-over-run.
 
 ### Degradation protocol (how LR is generated)
 
-Both datasets derive their LR input from HR via bicubic resizing, selectable per-dataset
-via `resize_backend`:
-
-- **`'matlab'` (default, tracked templates use it).** A vendored (not a dependency),
-  MATLAB-`imresize`-compatible bicubic resize — antialiased (MATLAB widens the
-  interpolation kernel by the scale factor on downscale; this widening *is* the
-  low-pass, so no separate blur step is used or accepted with this backend) and using
-  MATLAB's own bicubic coefficient (a=-0.5, vs. OpenCV's a=-0.75). This makes the LR
-  **inputs** directly comparable to published papers, most of which generate their
-  benchmark LR images with real MATLAB. See `sisr/imresize.py` for the implementation
-  and its license/attribution header, and `tests/test_imresize.py` for verification.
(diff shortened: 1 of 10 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #56, opened 2026-08-02
Quiet TensorBoard logging: wider cadences + per-image scalar gate

YeapJieShen:
## Why

A finished 1M-step SRResNet run wrote a **779 MB** TensorBoard event file. The CSV logger received the **identical scalar stream** and wrote **243 KB**. Scalars are therefore ~0.03% of the file; image strips (50 val cycles x 19 images per bicubic|SR|HR strip, ~820 KB per cycle) are essentially all of it.

BSD100 is now in fit-time validation (feeds both `val_dataloader` during fit and `test_dataloader`), taking strips from 19 to 119 per logged cycle. Per the standing decision, BSD100 stays in fit-time val -- this PR does not move it to test-only; ~1.3 GB event files per run are the accepted cost.

## Part A -- wider cadences (both templates)

- `trainer.val_check_interval`: 20000 -> 50000
- `GradNormLogger.log_every_n_steps`: 2000 -> 5000
- `trainer.log_every_n_steps`: 500 -> 1000
- `BenchmarkImageLogger.log_every_n_val_runs` stays 1 -- strips every val cycle; the reduction comes from the val cadence instead.

Not touched: `max_steps`/`batch_size` (owned by a separate stacked PR that raises `val_check_interval` further, to 500000, for the SRCNN template specifically).

**Costs, so they're chosen rather than discovered:** the val PSNR curve drops from 50 points to 20 over a 1M-step run; `SRCheckpoint` now selects top-k from 20 candidates instead of 50.

## Part B -- decouple per-image scalars from image strips

`BenchmarkImageLogger._flush_buffer` emitted per-image PSNR/SSIM scalars only inside the image-strip branch, welding two independent concerns together. At template defaults that's 19 images x (2 psnr + 2 ssim) = 76 TensorBoard series, 190 with `separate_psnr: true`, and 1190 with BSD100 + `separate_psnr`. The cost is **tag count**, not bytes (~285 KB) -- TensorBoard's scalar pane stops being navigable well before that.

Added a `log_per_image_metrics: bool = False` constructor flag, following the `WeightHistogramLogger` precedent (its default was raised 100 -> 10000 for the same reason: quiet default wins). Neither template gets an explicit entry for it. Per-set means (`psnr/{set}/{key}`, `ssim/{set}/{key}`) are unaffected and always logged regardless of this flag.

`cli test`'s `on_test_epoch_end` still passes `should_log_images=True` unconditionally -- correct for a one-shot final eval, unchanged by this PR.

## Test plan

- [x] New test: default (`log_per_image_metrics=False`) omits `per_image/...` scalars but per-set means and image strips still log.
- [x] New test: `log_per_image_metrics=True` emits `per_image/...` scalars even on a val run where `should_log_images` is False (proves the two concerns are independently gated).
- [x] Existing add_image/add_scalar regression test updated to pass `log_per_image_metrics=True` explicitly (previously relied on the now-changed default).
- [x] Both templates resolve: `sisr.cli fit --config templates/config.{srcnn,srresnet}.template.yaml --print_config`, exit 0, new values confirmed present (`val_check_interval: 50000`, `log_every_n_steps: 1000`, `GradNormLogger.log_every_n_steps: 5000`).
- [x] Full suite: 349 passed, 1 skipped (base branch: 347 passed, 1 skipped, plus the two new tests above).

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -44,9 +44,10 @@ class BenchmarkImageLogger(Callback):
     SR output is center-padded when smaller than HR (``padding='valid'``) so
     all three panels share the same spatial size.
 
-    Per-image PSNR/SSIM scalars go under ``"per_image/{name}/psnr/{key}/{filename}"``
-    and ``"per_image/{name}/ssim/{key}/{filename}"``; per-set means under
-    ``"psnr/{name}/{key}"`` and ``"ssim/{name}/{key}"``.
+    Per-set means go under ``"psnr/{name}/{key}"`` and ``"ssim/{name}/{key}"``
+    every cycle. Per-image scalars under ``"per_image/{name}/psnr/{key}/{filename}"``
+    and ``"per_image/{name}/ssim/{key}/{filename}"`` are gated by
+    ``log_per_image_metrics`` (default off — see that arg).
 
     Border cropping is sourced from ``pl_module.eval_config.crop_border`` at
     the use site; this avoids dual-knob configuration drift.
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #57, opened 2026-08-02
Add dormant byte-equality test for SRCNN's upscale degradation leg

YeapJieShen:
## Summary

`tests/test_imresize.py`'s byte-equality test only pairs HR with `LR_bicubic` — every case it covers is an HR-to-LR **downscale**. But `sisr/datasets/srcnn.py`'s `_degrade` does bicubic-down **then bicubic-up** (back to HR size, SRCNN's pre-upsampled formulation). The upscale leg has never had byte-equality evidence, only structural tests (kernel widening, constant-image invariance, dtype/clipping).

This PR adds the plumbing so an upscale byte-equality test activates automatically once the reference data exists:

- `_upscale_cases()`, mirroring the existing `_reference_cases()` collector, pairs each `LR_bicubic/X{s}/<stem>x{s}.png` with a `Bicubic_up/X{s}/<stem>x{s}.png` counterpart (same `("Set5", "Set14", "B100")` datasets tuple, same accumulation/reporting style).
- `test_matlab_imresize_upscale_matches_real_matlab_reference_data()` reads the trusted reference LR, upsamples it with `matlab_imresize` to `scale × lr.shape`, and asserts byte-equality against `Bicubic_up`.
- The module docstring documents the new `Bicubic_up/X{2,3,4}` layout, how to generate it in MATLAB (`imresize(lr, [h_crop w_crop], 'bicubic')`), and states the honest current limitation: only the downscale leg is byte-verified today.

**The reference data does not exist yet** — the user will generate it later in MATLAB Online. Until then this test **skips cleanly**, exactly like the existing downscale test does without `data/reference/`.

## Verification

Reference data is absent in this environment, so the real assertion can't run. Instead, both directions of the plumbing were proven:

1. **Skips cleanly today:**
   ```
   SKIPPED [1] tests\test_imresize.py:259: No Bicubic_up/ reference data under .../data/reference -- generate it per this file's module docstring to enable the upscale byte-equality check.
   ```
2. **Genuinely detects a wrong image** (not just skip-or-trivially-pass): a standalone fixture was built in a tmp dir with a real LR image and a *deliberately incorrect* `Bicubic_up` image (independent random noise, correct shape, not derived from `matlab_imresize` — so this isn't circular). Pointing `_upscale_cases` at that tmp dir and rerunning the exact mismatch loop found the mismatch, and the test's final `assert not mismatches` raised:
   ```
   1/1 reference images did not reproduce byte-exactly: ['Set5/babyx2 x2']
   ```
   A companion fixture with a *correctly* generated `Bicubic_up` image confirmed zero mismatches, so the check isn't vacuously failing either.
3. The existing 357-case downscale test is unchanged in behavior (same skip message/location, still skips).

## Test plan

- [x] `PYTHONPATH="$PWD" python -m pytest -q` → **347 passed, 2 skipped** (base branch: 347 passed + 1 skipped; +1 skip from the new dormant test)
- [x] `ruff check .` → all checks passed
- [x] `ruff format --check .` → 57 files already formatted
- [x] Manual fixture proving genuine mismatch detection (see above)

### DIFF
--- a/tests/test_imresize.py
+++ b/tests/test_imresize.py
@@ -6,7 +6,7 @@ Two layers of evidence:
    this a MATLAB-parity port rather than "generic bicubic": the a=-0.5
    kernel coefficient, antialiasing kernel widening on downscale, symmetric
    (mirror) border padding, and round-half-away-from-zero on the uint8 cast.
-2. A byte-equality test against real MATLAB-generated reference data —
+2. Byte-equality tests against real MATLAB-generated reference data —
    the strongest available claim, since without MATLAB itself this project
    cannot prove byte-identity from first principles.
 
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #60, opened 2026-08-02
Fix --ckpt_path failing to reload any checkpoint this project saves

YeapJieShen:
## Summary

- `SRLightning.__init__` overwrote `self._hparams` with a `/`-flattened dict built for clean TensorBoard HParams columns, so every saved checkpoint's `hyper_parameters` carried keys like `training_config/example_input_shape/0`. `LightningCLI._parse_ckpt_path` re-parses that dict verbatim as CLI options (`model.<key>`); jsonargparse rejects a literal `/` in an option name, so `fit` resume, `validate`, `test`, and `export` all failed on any `--ckpt_path`.
- `self.hparams` now holds only a plain, nested (`dataclasses.asdict`) `training_config`/`eval_config` dict, saved via an explicit `save_hyperparameters(...)` call so it's identical whether `SRLightning` is built directly or through the CLI, and safe under `weights_only=True` loading. The flattened, TensorBoard-only view moves to `self._tb_hparams`, which `on_train_start` now logs instead of `self.hparams`.
- Existing checkpoints (including a live 17-hour SRResNet run) were saved with the old flattened `hyper_parameters` and can't be regenerated, so `SRLightningCLI` also gets a `_parse_ckpt_path` override (`_reconstruct_ckpt_hparams`) that rebuilds the current nested shape from either format before reparsing — already-saved checkpoints keep loading too.

## Verification

- New regression tests in `tests/test_cli.py`: a real (non-`fast_dev_run`) one-step `fit` writes an actual checkpoint, then `validate`/`fit --ckpt_path` reload it through the CLI; a companion test rewrites that checkpoint's `hyper_parameters` into the legacy `/`-flattened shape and confirms it still reloads; unit tests for `_reconstruct_ckpt_hparams` directly.
- Ran the exact real-world repro against the project's live checkpoint (`experiments/SRResNet/20260802-paper-range/checkpoints/sr-940000-psnr_val_RGB=28.8497.ckpt`) via a scratch-copied config (`save_dir` redirected away from the real experiment directory) — it now completes a full Set5/Set14 test run instead of raising `SystemExit`, reproducing the project's documented benchmark numbers.
- Full suite: 368 passed, 1 skipped (pre-existing, no `data/` in this checkout). `ruff check` / `ruff format --check` clean.

## Test plan

- [x] `PYTHONPATH="$PWD" python -m pytest` — 368 passed, 1 skipped
- [x] `ruff check .` / `ruff format --check .` — clean
- [x] Reproduced the failure on unfixed code, then confirmed the fix resolves it against the real checkpoint
- [x] Confirmed no files were modified under `experiments/SRResNet/20260802-paper-range/`

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -31,7 +31,8 @@ already assigns ``torch.backends.cudnn.benchmark``, and it coordinates with
 """
 
 import sys
-from typing import Literal
+from pathlib import Path
+from typing import Any, Literal
 
 import torch
 from lightning.pytorch import Trainer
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #61, opened 2026-08-02
Add model IO declarations for checkpoint/export provenance metadata

YeapJieShen:
## Summary

This PR only adds declarations and changes no runtime behaviour. A later PR will consume these to write provenance metadata (`{scale, input, input_channels, output_range}`) into checkpoints, bare `.pt` weights, and ONNX exports — describing how to feed a model and what comes out, for anyone downloading weights without the training config alongside them.

- **`SRProcessor.output_range`** — new abstract property (mirroring the existing abstract `model_channels`), returning the model's output intensity range: `RGBProcessor`/`YCbCrProcessor`/`YChannelProcessor` → `(0.0, 1.0)`, `RGBSignedOutputProcessor` → `(-1.0, 1.0)`. Abstract rather than defaulted — a wrong inherited default is exactly the hazard this exists to prevent (applying `RGBProcessor`'s `[0, 1]` reconstruction to `RGBSignedOutputProcessor`-trained weights silently produces wrong images).
- **`SRModel.input_contract`** — declared explicitly per architecture as a `ClassVar[Literal["pre_upsampled", "native_lr"]]`, matching the file's existing style (`_hparams: dict` as a typed class-level attribute, alongside the `hparams` property). SRCNN → `"pre_upsampled"` (resolution-preserving; LR arrives already bicubic-upsampled to HR size). SRResNet → `"native_lr"` (consumes true low-resolution, upsamples internally). Deliberately not inferred from `'scale' in model.hparams` — that check happens to distinguish the two current architectures but would silently break on a third that doesn't follow the same pattern.
- **`SRTrainingConfig.scale`** — new optional `int | None` field (default `None`, so every existing config stays valid), cross-checked in `validate_against` against `model.hparams['scale']`: a mismatch raises `ValueError` when *both* sides are set; either side absent is a silent no-op. Not sniffed from `trainer.datamodule.train_dataset.scale` — that attribute lives outside the `SRDataset` contract (`PredictDataset` has none at all), whereas per-paper knobs already live in these config dataclasses.
  - `SRResNetTrainingConfig` sets `scale = 4` — Ledig et al. §3.2's SRResNet baseline is fixed at a single 4x factor, matching the model's own `scale` hparam, so the default is now actively validated rather than just recorded.
  - `SRCNNTrainingConfig` leaves `scale` at the inherited `None` — the model itself carries no `scale` hparam, and the paper reports x2/x3/x4 rather than one fixed value, so there is no single paper-correct default to pin.

## Test plan
- [x] `output_range` implemented and tested on all four processors; a bare `SRProcessor` subclass missing only `output_range` raises `TypeError` (abstractness confirmed alongside the existing `model_channels` case)
- [x] Both models' `input_contract` asserted (`SRCNN.input_contract == "pre_upsampled"`, `SRResNet.input_contract == "native_lr"`)
- [x] `validate_against` scale checks: raises on mismatch, no-ops when either side is absent (config unset, or model has no `scale` hparam) — at both the base `SRTrainingConfig` level and via `SRResNetTrainingConfig`'s new paper-fixed default
- [x] Both templates resolve: `sisr.cli fit --config templates/config.srresnet.template.yaml --print_config` (shows `scale: 4`) and the srcnn one (shows `scale: null`) — both exit 0
- [x] Full suite: **375 passed, 1 skipped** (base #54 was 363 passed + 1 skipped; +12 new tests, 0 regressions, 0 unaccounted-for changes)
- [x] `ruff check .` — all checks passed; `ruff format --check .` — 57 files already formatted

### DIFF
--- a/sisr/models/base.py
+++ b/sisr/models/base.py
@@ -1,6 +1,7 @@
 """Abstract base class for single-image super-resolution architectures."""
 
 import abc
+from typing import ClassVar, Literal
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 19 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #62, opened 2026-08-02
Unify SRCNN/SRResNet raw-HR caches; add advisory build lock

YeapJieShen:
## Summary

- Unifies SRCNN's and SRResNet's train-dataset HR caches into one architecture-neutral module (`sisr/datasets/hr_cache.py`): same cache name (`hr_raw`), same format tag (`hr_rgb_v1`), same headered value shape, one shared `process_hr_image`. The same image directory now produces exactly one on-disk cache regardless of which architecture builds it first, instead of two ~6.3 GB copies of identical pixels.
- Fixes an import-weight bug this uncovered: the function pickled to a spawned `ProcessPoolExecutor` build worker was `_process_hr_image` living in `srcnn.py`/`srresnet.py` (both of which import torch directly) — a spawned worker unpickles a function by re-importing its *defining* module, so every build worker was paying torch's full cold-import cost. Confirmed with a real spawned-worker probe before (`torch` present) and after (`torch` absent); guarded by a new test on the worker entry point's actual module.
- Adds an advisory file lock (`O_CREAT|O_EXCL` sentinel) so two processes racing to build the same `cache_dir` don't duplicate the decode pass. A crashed builder's stale lock cannot deadlock a later run — a loser gives up after `lock_timeout` and builds anyway.
- Fixes a latent bug in `LMDBCache._try_load` surfaced by testing the lock against real concurrent processes: a `lock=False` reader can raise mid-read (not just fail to open) while a genuine writer in another process is still active, and the old code treated that identically to corruption, deleting the directory out from under the other process's in-progress build.

## Diagnosis: torch in the build worker (Task B)

Reproduced with a real `ProcessPoolExecutor` (not a guess): submitted the actual `_process_hr_image` reference from `sisr.datasets.srcnn`, then a second job in the same worker checking `sys.modules`.

- Before: `torch in worker sys.modules after _process_hr_image ran: True`
- After (via the new `sisr.datasets.hr_cache` module, both directly and through each dataset's re-exported alias): `False` in all three cases

## Cache sharing / pixel equivalence

New tests in `tests/datasets/test_hr_cache.py`:
- Build via SRResNet, then construct SRCNN over the same files — cache path and checksum match, `_process_hr_image` not called (no rebuild). Same in the other build order.
- Sub-images/crops read back through the shared cache by either architecture are byte-identical to the raw source pixels (SRCNN's fixed corner patch and SRResNet's random crop, found verbatim in the source).

## Concurrency

- Kept the lock in `sisr/cache.py` (stdlib only, torch-free), scoped per `(cache_dir, name, checksum)`.
- Two real OS processes racing to build the identical cache: `build_fn` runs exactly once; both processes end up with the correct value (`tests/test_cache.py::test_lmdb_cache_builds_exactly_once_across_two_racing_processes`). A stale lock with no valid cache behind it times out and builds anyway rather than blocking forever.
- Checked whether pytest-xdist workers already hit this race: no — each xdist worker gets its own `popen-gwN` temp root (verified empirically), so `tmp_path`/`tmp_path_factory` fixtures never collide across workers even with identical basenames, regardless of `--dist=loadscope`.

## Test plan

- [x] `PYTHONPATH="$PWD" python -m pytest -q` — base branch (`chore/drop-cv2-backend`) is 347 passed + 1 skipped; this branch is **360 passed + 1 skipped** (13 new tests, no regressions), also verified under `-n 2 --dist=loadscope` (CI's exact invocation).
- [x] Cache-unification and pixel-equivalence tests pass, including cross-architecture reuse in both build orders.
- [x] Lock tests pass repeatedly (8 consecutive runs) including the real dual-process race.

### DIFF
--- a/sisr/cache.py
+++ b/sisr/cache.py
@@ -16,6 +16,7 @@ torch-free; colorspace math lives in :mod:`sisr.colorspace` for this reason.
 
 import os
 import shutil
+import time
 from collections.abc import Callable, Iterator, Sequence
 from concurrent.futures import FIRST_COMPLETED, Future, ProcessPoolExecutor, wait
 from contextlib import contextmanager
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #64, opened 2026-08-02
Add distributable weights checkpoint + provenance metadata

YeapJieShen:
## Summary

- One `build_metadata()` function (`sisr/training/metadata.py`) produces a single provenance dict — model `class_path`/`init_args`, processor identity, IO contract (scale, input contract, channels, output range), eval config, training state — consumed by all three sinks below, so they cannot drift apart.
- `SRLightning.on_save_checkpoint` injects a `sisr_meta` key into every saved `.ckpt` via the public Lightning hook. `LightningCLI._parse_ckpt_path` only reads `hyper_parameters`, so this is inert for `--ckpt_path` resumption.
- New `SRWeightsCheckpoint` (`sisr/training/callbacks.py`), a `ModelCheckpoint` sibling to `SRCheckpoint`, overrides the private `_save_checkpoint` hook to write a bare `.pt` — `pl_module.model.state_dict()` (no `model.` prefix) + `meta` — instead of a full training checkpoint. Distinct `FILE_EXTENSION`/`filename_prefix` so it coexists with `SRCheckpoint` in one `dirpath` without cross-deleting files. Wired into both templates, enabled by default.
- `sisr/export.py` now stamps the same metadata onto exported ONNX graphs as per-field `metadata_props` (JSON-encoded per field, so Netron renders a readable table), carrying a source checkpoint's own provenance forward when `ckpt_path` is given.

**Measured size reduction:** a real SRResNet checkpoint (1,549,462 params, Adam) — `.ckpt` 18.83 MB vs `.pt` 6.31 MB, **~3x smaller**, matching the ~18.8 MB / 3x figures in the task description.

`processor` (class_path) and `io.output_range` are what make a bare `.pt` safe to hand out on its own: `RGBSignedOutputProcessor` trains in `[-1, 1]` HR/output range while `RGBProcessor` uses `[0, 1]` — applying the wrong reconstruction to a bare state_dict produces silently wrong images with no error, so recording it in the metadata means a downstream consumer never has to guess.

## Test plan
- [x] `PYTHONPATH="$PWD" python -m pytest -q` → **404 passed, 1 skipped** (base was 375 passed + 1 skipped; +29 new tests)
- [x] `ruff check .` / `ruff format --check .` clean
- [x] Both templates resolve via `--print_config`, with `SRWeightsCheckpoint` visible alongside `SRCheckpoint`
- [x] Dedicated end-to-end `Trainer.fit` test proves: the private `_save_checkpoint` hook is actually invoked (loud-failure guard), the `.pt` has no optimizer state / no `model.`-prefixed keys, the state_dict loads `strict=True` into a fresh bare model, and `SRCheckpoint`/`SRWeightsCheckpoint` don't delete each other's files sharing one `dirpath`
- [x] ONNX `metadata_props` test asserts the fields are present and parse back to expected values; skips cleanly without the `[export]` extra

### DIFF
--- a/sisr/export.py
+++ b/sisr/export.py
@@ -8,6 +8,12 @@ oversight: a consumer of the exported graph is expected to have ``sisr``
 importable and call ``processor.extract`` / ``processor.reconstruct``
 themselves.
 
+The exported graph also carries the same provenance metadata as the checkpoint
+sinks (:func:`sisr.training.metadata.build_metadata`), written one field per
+``onnx.ModelProto.metadata_props`` entry rather than a single opaque JSON blob —
+Netron renders per-field props as a readable table, which is most of the value.
+``metadata_props`` is string->string only, so non-string fields are JSON-encoded.
+
 For :class:`~sisr.processors.RGBProcessor` architectures (SRResNet), those
 methods are identity functions, so the bare graph is already an end-to-end
 LR-RGB -> SR-RGB pipeline. Under
@@ -27,6 +33,7 @@ actually invoked.
 
 from __future__ import annotations
 
+import json
 import warnings
 from os import PathLike
 from typing import TYPE_CHECKING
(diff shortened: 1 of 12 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #65, opened 2026-08-02
Remove internal tracker identifiers from tracked files

YeapJieShen:
## Summary

- Removed 69 internal tracker-identifier occurrences (`INIT.xx`, `Px.y`, and the word "triage") across 24 tracked files — private notes-directory vocabulary that means nothing to a public reader. PR numbers like `#53` were left untouched; they resolve on GitHub.
- The surrounding rationale was preserved in every case — only the bracketed/inline codes were deleted, never the explanatory sentence.
- Three spots needed a rewrite rather than a deletion: `tests/test_colorspace.py` (dropped the "triage P2.8" internal-document reference, rewrote the sentence to stand on its own), `tests/training/test_config.py` (replaced the delivery-wave label `B1/C1` with the actual consumers, `SRLightning` and `BenchmarkImageLogger`), and `.gitignore:8` was intentionally left alone (already agreed to keep as-is).

## Verification

Both required greps come back empty (or with only the expected exception):

```
$ git ls-files | xargs grep -nE "\b(INIT\.[0-9]+|[Pp][1-5]\.[0-9]+)\b"
(no output)

$ git ls-files | xargs grep -nEi "triage|wave [a-z]\b|\bB1/C1\b"
.gitignore:8:# Claude's local working notes (codebase index, triage, session history)
```

No-behavior-change proof: parsed every touched `.py` file (20 files) at both `main` and this branch, stripped docstring nodes, and compared `ast.dump()` — all 20 matched exactly.

- Full suite: `423 passed, 2 skipped in 34.66s` (matches the `main`-worktree baseline; the 2 skips are the reference-data-absent imresize checks, expected in a worktree with no `data/`).
- `ruff check .` → All checks passed. `ruff format --check .` → 61 files already formatted.
- Both templates resolve: `sisr.cli fit --config templates/config.srcnn.template.yaml --print_config` and the srresnet equivalent both exit 0.

## Test plan

- [x] `git ls-files | xargs grep -nE "\b(INIT\.[0-9]+|[Pp][1-5]\.[0-9]+)\b"` empty
- [x] `git ls-files | xargs grep -nEi "triage|wave [a-z]\b|\bB1/C1\b"` only `.gitignore:8`
- [x] AST-equivalence (docstrings stripped) across all 20 changed `.py` files
- [x] `pytest -q` → 423 passed, 2 skipped
- [x] `ruff check .` / `ruff format --check .` clean
- [x] Both templates' `--print_config` exit 0

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,4 @@
-# github-actions only (INIT.6): a PEP 621 pyproject with `>=` lower bounds and no
+# github-actions only: a PEP 621 pyproject with `>=` lower bounds and no
 # lock file gives the pip updater almost nothing to do, and enabling it would need
 # an `ignore` entry for albumentationsx (capped at <=2.1.0 for a documented Windows
 # import failure, see pyproject.toml) that a later PR is about to make obsolete anyway.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Smoke import
         run: python -c "import sisr; from sisr.cli import main; print('sisr', sisr.__version__, '- main:', main)"
       # Ports the old tests/test_cli.py::test_cli_help_lists_subcommands subprocess
-      # smoke here (P5.11): it cost ~10s of the pytest suite re-importing torch +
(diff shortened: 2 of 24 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #67, opened 2026-08-03
Raise torch floor past the weights_only bypass and harden workflow tokens

YeapJieShen:
## What

Two related hardening concerns, three commits:

1. **`torch>=2.6`, `torchvision>=0.21`** — `torch.load(..., weights_only=True)` is this project's entire load-time safety contract (`sisr/cli.py`, `sisr/export.py`, `sisr/training/lightning_module.py`), and the project deliberately ships distributable `.pt`/`.ckpt` files. GHSA-53q9-r3pm-6pq6 / CVE-2025-32434 makes that flag bypassable for arbitrary code execution on torch ≤ 2.5.1 (fixed in 2.6.0), and a floor-only spec never upgrades a pre-existing environment — so the floor itself must exclude the vulnerable range. A new `tests/test_packaging.py` invariant (tomllib parse, never imports `sisr`) guards both floors.
2. **Workflow token hardening** — explicit `permissions:` blocks in both workflows (`contents: read`, plus `actions: write` with an inline comment: an explicit block zeroes unlisted scopes, and Actions-cache *saves* need `actions: write` — without it the setup-uv/setup-python caches silently stop saving). `codecov/codecov-action` repinned from the mutable `@v7` tag to the resolved commit SHA (verified to be the genuine `v7.0.0`). `astral-sh/setup-uv@v9.0.0` left as-is — already deliberately patch-pinned.

## Evidence (test proven RED → GREEN)

RED, before the floor bump:
```
tests/test_packaging.py::test_torch_floor_excludes_weights_only_bypass FAILED
E   AssertionError: torch floor must be >= 2.6: ['torch>=2.4', 'torchvision>=0.19', ...]
E   assert (2, 4) >= (2, 6)
```
GREEN, after:
```
tests/test_packaging.py::test_torch_floor_excludes_weights_only_bypass PASSED
```

Full suite from the branch: **424 passed, 2 skipped** (the two data-dependent imresize byte-equality tests, which skip without local reference data by design); `ruff check` + `ruff format --check` clean. Both workflow files re-validated with `yaml.safe_load` after each change.

Adversarially reviewed (spec compliance ✅ / quality approved), including independent re-verification that the CVE is real and fixed in 2.6.0, that the pinned SHA resolves to the genuine `codecov-action` v7.0.0, and that the packaging test's version comparison is tuple-based (no lexicographic-string trap).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,9 @@ on:
 
 permissions:
   contents: read
+  # An explicit permissions block zeroes unlisted scopes; actions: write keeps
+  # Actions-cache saves working (setup-uv enable-cache / setup-python cache: pip).
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,9 @@ on:
     branches: [main]
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ on:
 
 permissions:
   contents: read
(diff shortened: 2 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #68, opened 2026-08-03
Extract shared HR-cache wiring into a dataset base class

YeapJieShen:
## What

Both training datasets copy-pasted the wiring around `hr_cache.py`'s primitives — checksum delegation, the PIL size probe, the ~10-line `LMDBCache` construction, the header-decode dance in `__getitem__` — and the copies had already drifted (two different tqdm labels for the *same shared cache*). This extracts an `HRCachedTrainDataset` base in `sisr/datasets/base.py`; each architecture now contributes only its indexing scheme (grid vs random crop) and its degradation.

## Behaviour-neutrality (the acceptance criterion)

- `sisr/datasets/hr_cache.py`: **zero-line diff** — the torch-free pool-worker boundary is untouched (guard test still targets it).
- Checksum inputs, cache directory naming, `FORMAT_TAG`, stored byte format, and build parallelism: all verified unchanged (adversarial review traced each against the pre-change code).
- New regression test: an existing on-disk cache re-opens through **both** architectures against the same directory with the build path proven never invoked (`process_fn` mock asserted not called — the actual mechanism, not a timing proxy).
- Public constructor signatures and YAML `class_path`s unchanged; existing tests needed zero adaptation.
- Suite: **424 passed, 2 skipped** (data-dependent imresize tests skip in a worktree by design); `ruff check` + `ruff format --check` clean.

LOC: `srcnn.py` −63, `srresnet.py` −62, `base.py` +126 (mostly the mandatory Google-style docstrings on the new public base) — the duplication is gone even though net LOC lands flat.

Adversarially reviewed: spec ✅, quality approved, no Critical/Important findings (three minor docstring polish notes deferred to the batch's final whole-branch review).

### DIFF
--- a/sisr/datasets/base.py
+++ b/sisr/datasets/base.py
@@ -6,14 +6,24 @@ glob + empty-directory guard), the RGB image load, the ``uint8`` HWC →
 ``float32`` CHW ``[0, 1]`` tensor adapter, and declares the ``.img_paths``
 filename contract that :class:`~sisr.training.SRDataModule` and
 :class:`~sisr.training.callbacks.BenchmarkImageLogger` rely on.
+
+:class:`HRCachedTrainDataset` also centralises the raw-HR LMDB wiring shared
+verbatim by SRCNN's and SRResNet's train datasets — it lives here, not in the
+torch-free :mod:`~sisr.datasets.hr_cache`, because it subclasses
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #69, opened 2026-08-03
Speed up the imresize hot path and benchmark-callback metrics

YeapJieShen:
## What

Three measured performance fixes from the four-dimension audit, byte-exactness preserved and proven:

1. **`_contributions` memoized** (`functools.lru_cache`, arrays frozen read-only) — it is a pure function of `(in_length, out_length, scale)` and constant per configured dataset, yet was recomputed on every resize call (measured at 48% of SRCNN's per-sample degrade, 20% of SRResNet's resize).
2. **`_resize_axis` single-pass `einsum` contraction** — the old multiply-then-`np.sum` materialized a second full float64 temporary (~265 MB per axis on a DIV2K validation image; >530 MB transient per call).
3. **Benchmark callback**: PSNR/SSIM now computed from the on-device slices *before* any host copy (mirroring what the primary val path already did), exactly one host transfer per image and only on image-logging cycles, image strips emitted immediately instead of buffering ~0.5 GB of full-res tensors per validation cycle, and the buffer is a `BenchmarkSample` NamedTuple (killing the positional `s[4]`/`s[5]` reads).

## Evidence

- **Byte-exactness (the non-negotiable):** the full **357-case MATLAB reference corpus passes against this exact head**, verified independently from the main checkout with this branch's code injected (`-P` + `PYTHONPATH`, module resolution asserted). Unit-level equivalence tests: einsum vs a naive multiply-sum oracle byte-identical across non-square/1-channel/3-channel shapes; cached vs `__wrapped__` output identical; read-only enforcement tested by actual mutation attempt.
- **Metric-value equivalence:** a dedicated test reconstructs the pre-change ordering (host-copy → crop → metrics) against the shipped ordering on the same tensors — every PSNR/SSIM key equal at `abs=1e-6`, CPU-only (CI-safe).
- **TensorBoard tags byte-identical** (verified string-by-string in review), per-set means unchanged.
- **Micro-benchmarks (this machine; directional — shared hardware):** SRCNN per-sample degrade path 572 → 319 µs/call (**1.79×**); SRResNet resize path 1545 → 837 µs/call (**1.85×**); callback metric computation **6.09× / 3.54×** faster by avoiding the pre-metric host copy. Audit-time baselines measured the same shapes at 2.21×/1.41×/10.6× on quiet hardware; final serial numbers land with the batch close-out.
- Suite: **434 passed, 2 skipped** (data-dependent imresize tests skip in a worktree by design); ruff clean.

Adversarially reviewed twice (initial review verified the einsum subscripts by re-derivation and re-ran the affected test files; verdict Approved after the equivalence-test fix).

### DIFF
--- a/sisr/imresize.py
+++ b/sisr/imresize.py
@@ -60,6 +60,7 @@ SOFTWARE.
 ------------------------------------------------------------------------------
 """
 
+import functools
 from math import ceil
 
 import numpy as np
@@ -84,6 +85,7 @@ def _round_half_away_from_zero(x: np.ndarray) -> np.ndarray:
     return np.sign(x) * np.floor(np.abs(x) + 0.5)
 
 
+@functools.lru_cache(maxsize=32)
 def _contributions(in_length: int, out_length: int, scale: float) -> tuple[np.ndarray, np.ndarray]:
     """Per-output-pixel interpolation weights and source indices for one axis.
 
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #70, opened 2026-08-03
Enforce the model-dataset contract and make config mistakes fail loudly

YeapJieShen:
## What

Closes the one configuration cross-wire in the framework that corrupted results **without any error** — plus two adjacent validation gaps. Three concerns, six commits:

1. **`input_contract` is now enforced.** Previously its only consumer was provenance metadata: pairing SRResNet with the SRCNN dataset in YAML made `center_crop` silently zero-pad HR to 4× (93.75% of the loss target became black border — loss still decreased, PSNR still reported). A `setup()`-time probe now samples one `(lr, hr)` pair via the datamodule and validates it against the model's declared contract (`pre_upsampled` ⇒ equal spatial dims; `native_lr` ⇒ `hr == lr × scale`, falling back to the model's own `scale` hparam when the config leaves it unset), raising an actionable error naming the model, contract, dataset class, both shapes, and two remedies. A belt-and-braces guard in `_forward_sr` refuses to crop when HR is smaller than SR. Verified live: the probe **cannot false-positive on any shipped configuration** (both validation datasets guarantee exact LR/HR relationships by construction).
2. **`example_input_shape` H/W validated** against the real sampled train patch (it drives the compile warm-up shape and FLOPs; only the channel count was checked before).
3. **Dataset CLI overrides fail loudly instead of silently.** `--data.train_dataset.init_args.x=y` used to parse cleanly and do *nothing*; whole-dict JSON forms crashed with a bare jsonargparse `AttributeError`. Now: malformed specs are rejected with actionable messages at construction, and the parse-time crash is converted to an actionable error **only** when it is genuinely that collision — whole-dict overrides of fields with no prior value (the documented `sisr predict` workflow) keep working, byte-for-byte verified against the pre-change baseline.

## Evidence

- All gaps proven **RED → GREEN**: cross-wired pairing (both directions) raises at setup where it previously trained silently; wrong `example_input_shape` raises where it previously passed; each malformed CLI form now errors actionably where it previously no-op'd or crashed raw. RED outputs captured against the pre-fix tree via stash/pop.
- A real `Trainer(fast_dev_run=True)` integration test exercises the probe end-to-end (guards the Lightning hook-ordering premise).
- The probe is deterministic-safe: single sample, global RNG state snapshot/restored (seeded crop sequences unchanged), no LMDB build triggered, graceful skip on foreign datamodules.
- Suite: **455 passed, 2 skipped** (33 new tests); ruff clean; templates parse unchanged.

Adversarially reviewed across three rounds (Opus): initial review attacked the probe for false positives on shipped configs (none constructible), round 2 caught a guard over-rejection that would have regressed the predict workflow — fixed and byte-for-byte verified — and the final round approved with the remaining minors explicitly recorded.

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -30,13 +30,14 @@ already assigns ``torch.backends.cudnn.benchmark``, and it coordinates with
 ``trainer.deterministic``. A separate top-level key would silently bypass that.
 """
 
+import re
 import sys
 from pathlib import Path
 from typing import Any, Literal
 
 import torch
 from lightning.pytorch import Trainer
-from lightning.pytorch.cli import LightningCLI
+from lightning.pytorch.cli import ArgsType, LightningArgumentParser, LightningCLI
 
 from .export import to_onnx
 from .training import SRDataModule, SRLightning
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #71, opened 2026-08-03
Make the LMDB cache build lock unable to destroy live data

YeapJieShen:
## What

The cache build lock had a destructive fall-through: past `lock_timeout` (600 s — exceedable by a real DIV2K build on a slow disk), a second process would `rmtree` the live builder's memory-mapped LMDB (crash on Windows; on Linux the deletion *succeeds silently* and the builder spends the rest of its run writing to deleted inodes), then unlink a sentinel it never created, inviting a third process into the same race. Separately, `_try_load` treated *any* open failure as corruption and deleted the cache — including `lmdb.Error: environment already open in this process`, which is reachable through shipped code.

The fix, two commits:

- **Builds are atomic and never destructive.** Build into a pid-suffixed temp sibling; publish via rename-aside (`os.replace` to a trash sibling) + `os.replace` into place — no partial-deletion window exists at all. A target blocked by a reader's open handle produces a clear chained error with both directories intact, never a deletion.
- **Environmental failures raise; only genuine corruption rebuilds.** Deletion is now reserved for `lmdb.CorruptedError`/`InvalidError`/`VersionMismatchError`; bare `lmdb.Error`, `LockError`, `DiskError`, and `OSError`/`PermissionError` return "not ready" without touching disk, and builder-side open failures raise an actionable chained error instead of silently rebuilding for minutes. Missing metadata keys classify as stale instead of crashing.
- **A live builder is never preempted.** The sentinel carries pid + an mtime heartbeat (which retries transient errors); takeover requires dead-pid *and* staleness, with the pid re-read immediately before unlink; a sentinel recording our own pid is treated as dead (Windows pid recycling); waiting is bounded (3× `lock_timeout` → `TimeoutError` naming the sentinel and remedy, with periodic re-logging); release verifies the sentinel still records our pid. Windows pid-probing avoids `os.kill(pid, 0)` (which would *terminate* the probed process) and distinguishes access-denied (alive) from invalid-parameter (dead).
- **No disk leaks:** cache-sized orphan temp/trash siblings from crashed builds are swept under the lock once their pid is dead.

## Evidence (all RED on the pre-fix code)

- Live-builder-past-timeout race (two real processes): the loser neither deletes the live DB nor removes the winner's sentinel, and a shared build log asserts the build ran **exactly once** — platform-independent (the original failure was masked on POSIX, where the deletion succeeds silently; only a Windows `PermissionError` ever made it visible).
- Same-process double-open: construction now raises an actionable error instead of a silent full rebuild ending in deletion under a live reader.
- Ownership: a process that never acquired the sentinel provably leaves a foreign sentinel intact.
- The pre-existing test that *encoded* the bug (generic `lmdb.Error` ⇒ assert deleted) is tightened to genuine corruption.
- Suite: **441 passed, 2 skipped**; ruff clean; `sisr/cache.py` remains torch-free (guard test intact).

Adversarially reviewed twice (Opus, interleaving-level analysis both rounds): round 1 found six Important gaps including the POSIX-silent deletion and an unbounded self-pid wait — all verified closed in round 2; six non-blocking minors recorded for the batch's final review.

### DIFF
--- a/sisr/cache.py
+++ b/sisr/cache.py
@@ -14,8 +14,11 @@ build path never calls. ``tests/test_cache.py`` asserts the module stays
 torch-free; colorspace math lives in :mod:`sisr.colorspace` for this reason.
 """
 
+import logging
 import os
 import shutil
+import sys
+import threading
 import time
 from collections.abc import Callable, Iterator, Sequence
 from concurrent.futures import FIRST_COMPLETED, Future, ProcessPoolExecutor, wait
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #72, opened 2026-08-04
Skip training-path reconstruct and correct the templates' loader settings

YeapJieShen:
## What

Stacked on #70 (merge that first, then retarget this to `main`). Two things survived rigorous measurement; one thing was shipped, refuted by better measurement, and reverted *within this PR* — the history is kept honest below.

1. **`reconstruct` is no longer computed on the training path.** `training_step` always discarded it (`loss, *_`), yet every step paid for it — measured at +0.92 ± 0.26 ms/step on SRCNN and ~0 on SRResNet. Threaded as an explicit `need_sr_rgb` parameter (deliberately not `self.training`, which would break direct `predict_rgb` calls on un-`.eval()`'d modules); validation/test/predict/benchmark paths provably keep the full pipeline (call-site audit, spy tests on the real processor method, real-`Trainer` integration test). The crop target derives from the model output — all four shipped processors preserve H/W (independently re-verified).
2. **Templates: honest comments.** The bf16 comment no longer calls bf16-mixed "a bigger throughput lever" — measured, it is a 0.72× *regression* at the paper's batch 16 (crossover exists at b32). The `num_workers` comment now documents the real measured basis (below).

## The reverted middle commit (kept for honest history)

`af15524` shipped `num_workers: 0` based on steady-state figures that later proved to be a **synthetic-vs-real apples-to-oranges comparison** (the w0 side was measured in a worktree with no `data/`). The batch's serial benchmark pass on real DIV2K at Best-performance power showed w0 is a severe regression — SRCNN **79.3 ms/step vs 14.9** at w16 (batch 64 × ~1.08 ms/sample of real LR-derivation is a hard serial floor), SRResNet 56.0 vs 40.1 — and `43b138b` restores `num_workers: 16` / `persistent_workers: true` with the corrected comment. The ~2-minute one-time worker-spawn cost w0 saves would have traded against ~+179 hours on SRCNN's full training budget.

## Evidence

- Reconstruct skip: spy tests both directions, loss bit-identity, full-`Trainer` integration spy; branch code measured **neutral at w16** on real data (1.006×/0.972× vs main) — this commit's value is enabling the CUDA-graphs win in #74 (captured region is the skip-path shape).
- Templates parse via `--print_config` (re-verified after the revert); suite **463 passed, 2 skipped**; ruff clean.
- Reviewed: task review Approved (0 findings); the num_workers revert follows the controller's serial real-data benchmark (provenance-asserted per run, power state verified 115 W).

## Also evaluated and not shipped

A threaded in-process prefetch loader passed a bit-exact determinism gate but measured slower than stock in steady state; it is archived unpushed with a full writeup (its comparison numbers carry the same synthetic-data caveat, recorded for any future reopen).

### DIFF
--- a/sisr/training/lightning_module.py
+++ b/sisr/training/lightning_module.py
@@ -486,8 +486,10 @@ class SRLightning(lightning.LightningModule):
         """
         return self.model(x)
 
-    def _forward_lr(self, lr_img: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
-        """LR-only core of the forward pipeline: extract → model → reconstruct.
+    def _forward_lr(
+        self, lr_img: torch.Tensor, need_sr_rgb: bool = True
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        """LR-only core of the forward pipeline: extract → model → (reconstruct).
 
         Factored out of :meth:`_forward_sr` so :meth:`predict_step` can share
         the exact colorspace pipeline instead of forking it — HR is only ever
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #73, opened 2026-08-04
Correct every audit-verified false docstring claim

YeapJieShen:
## What

Doc-only commit fixing the four documentation claims that this batch's adversarial audit proved false against the code (stacked on #72 → #70; merge those first, then retarget this to `main`):

1. **SRCNN output-shape claim, three sites** — `SRCNN.forward`, `SRLightning.forward`, and `predict_rgb` all claimed output "same shape as *x*" while the shipped config (`kernel_sizes: [9,1,5]`, `padding: 0`) shrinks 33×33 → 21×21; the codebase's own crop/pad logic exists *because* the claim was false. Now stated padding-conditionally with the per-conv arithmetic, cross-referenced so the three sites can't drift apart again.
2. **`SRLightning.forward` claimed to be the `trainer.predict` path with no colorspace conversion** — it is bypassed entirely (`predict_step → _forward_lr` runs the full pipeline). Now names its real callers (TensorBoard `log_graph` tracing `example_input_array`, direct `module(x)`) and the actual routing — verified accurate against this stack's newest code (reconstruct skip, setup probe).
3. **Stale "blur" reference** in `sisr/datasets/srresnet.py` — the contrasted SRCNN degradation has no blur step (imresize's kernel widening *is* the low-pass); the repo-wide sweep confirmed this was the only stale remnant.
4. **`to_onnx` called `example_input_shape` "only a TensorBoard-graph dummy size"** — contradicted by its own function's default trace input 40 lines below, plus ModelSummary FLOPs and the compile warm-up. Now enumerates the real consumers accurately.

## Evidence

Each item was adversarially verified twice: once by the audit (verdict + exact wording), once by this PR's reviewer re-deriving every claim against the current branch (padding arithmetic, call-graph traces, config values). Diff is strictly docstring/comment prose — zero executable hunks. Suite independently re-run by the reviewer: **463 passed, 2 skipped**; ruff clean.

### DIFF
--- a/sisr/datasets/srresnet.py
+++ b/sisr/datasets/srresnet.py
@@ -42,7 +42,7 @@ class TrainDataset(HRCachedTrainDataset):
     never the crop), and bicubic-downsamples it by ``scale`` (via
     :func:`sisr.imresize.resize`) to form the LR input. Unlike
     :class:`sisr.datasets.srcnn.TrainDataset` there is **no
-    blur+downsample+upsample round-trip** and the LR is *not* upsampled back —
+    downsample+upsample round-trip** and the LR is *not* upsampled back —
     the model is responsible for the ×``scale`` upsampling, so the LR tensor is
     ``hr_crop_size // scale`` on a side.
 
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #74, opened 2026-08-04
Add opt-in full-step CUDA-graph capture for training

YeapJieShen:
## What

Opt-in CUDA-graph capture of the full training step (stacked on #72 → #70; merge those first, then retarget). `SRTrainingConfig.cuda_graph` (default **off**) captures `{zero_grad, extract, model, crop, loss, backward}` into one graph — the optimizer step deliberately stays eager, which is what keeps LR schedulers, `global_step`, grad clipping, and checkpointing working untouched.

**Why:** with the loader myths dispelled, SRCNN's training step is CPU-launch-bound — 78 kernel launches per step with the GPU at 27% utilization. Graph replay collapses them to one launch: measured through the real config surface (the only change being the flag), **5.91 → 2.23 ms/step = 2.65×, with losses bit-identical to 17 digits across interleaved runs**, GPU utilization 27% → 99%. On SRCNN's 10⁷-step budget that is on the order of ten hours saved per full run.

## Safety engineering (the interesting part)

- **Partial batches**: shape-mismatched batches take a bit-exact eager fallback, gated **per step** — a naive per-run gate demonstrably steps the optimizer on stale gradients at every epoch's final partial batch (RED-proven: loss diverges at exactly the first post-partial step).
- **Hard refusals with actionable messages**: mixed/low precision, `compile_backend`, DDP, gradient accumulation — the last re-checked **every epoch**, because `GradientAccumulationScheduler` mutates the value after fit-start and is *guaranteed* to bypass a fit-start-only check (review-caught Critical, RED-proven with the real scheduler).
- **Refit-safe**: graph state resets unconditionally at fit start — a stale graph after Lightning's teardown (which reallocates params/grads on CPU) would otherwise replay into freed memory, and a stale fallback flag would silently skip backprop on a flag-off refit.
- **Warm-up leaves the model bit-identical**: all buffers (BatchNorm running stats included) snapshot/restore around the warm-up forwards — proven on SRResNet, whose BN buffers otherwise drift invisibly to SRCNN-only parity tests.
- **Diagnosable**: one-shot capture log; a warning after persistent consecutive fallbacks so a silent 0%-speedup misconfiguration is visible.
- `lr_scheduler` deliberately **not** refused: the eager optimizer re-reads `param_groups[i]['lr']` every step, so schedulers are correct under this design — verified with a StepLR parity-and-decay test that trips if anyone ever moves the optimizer into the capture.

## Evidence

- Bit-exact parity: 200-step loss traces, exact equality (GPU-gated tests; CI-safe skips — CPU-only run: 483 passed / 8 skipped).
- Suite: **489 passed, 2 skipped**; ruff clean. 26 new tests, 13+ CPU-runnable (refusals, gating, reset, buffer snapshot all guarded without a GPU).
- Reviewed by an Opus reviewer in two rounds of interleaving-level analysis (every Lightning-internals assumption verified against the installed 2.13); one Critical and two Importants found, fixed, and RED-proven; final verdict Approved.

Definitive real-data numbers at Best-performance power state land in a follow-up comment from the batch's serial benchmark pass.

### DIFF
--- a/sisr/training/config.py
+++ b/sisr/training/config.py
@@ -130,16 +130,17 @@ class SRTrainingConfig:
             ``optimizer.step()`` stays eager, so LR schedulers, gradient
             clipping and ``global_step`` accounting are unaffected. Requires a
             CUDA device, ``precision='32-true'``, a single process, and
-            ``accumulate_grad_batches=1``; ``SRLightning.on_fit_start`` refuses
-            the rest rather than silently mistraining. Mutually exclusive with
-            ``compile_backend`` (see :meth:`__post_init__`). Validation, test
-            and predict always stay eager — their image sizes vary, and graphs
-            need static shapes. Measured on one RTX 5060 Laptop (SRCNN, batch
-            64, 33x33 Y patches, 60 W cap): **2.81x steps/s**, 6.21 -> 2.21
-            ms/step, bit-identical losses. The win is proportional to how
-            launch-bound the architecture is, so it is far smaller for
-            SRResNet, whose GPU floor is real compute — hence opt-in, per
-            config, defaulting off.
+            ``accumulate_grad_batches=1``; ``SRLightning.on_fit_start`` and
+            ``on_train_epoch_start`` refuse the rest rather than silently
+            mistraining. Mutually exclusive with ``compile_backend`` (see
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #76, opened 2026-08-05
Add pluggable losses: Charbonnier, total variation, and VGG16/19 perceptual

YeapJieShen:
Adds pluggable loss functions selectable and combinable entirely from YAML: Charbonnier, isotropic total variation, and VGG16/VGG19 feature-space perceptual losses, plus a named weighted sum that composes them. L1 and MSE need no new class — `torch.nn.L1Loss` already works as a criterion — so they ship as documented recipes instead of wrappers.

## ⛔ Do not merge yet

Seven `@requires_cuda` tests are currently **skipped**, not passing: a training run owns the machine's only GPU, so they were deliberately held. One of them is the sole real-hardware guard on the CUDA-graph interaction below. Suite is **612 passed / 9 skipped** (2 data-dependent + 7 GPU-held). Merge after those seven run green.

## The design

`SRLightning.criterion` was already YAML-wired with an `MSELoss` default, so the gap was the losses themselves plus a way for a loss to learn the model's output space. That is the one new seam: an `SRLoss` base declaring **abstract** `bind(processor)`, called once in `SRLightning.__init__`. A loss reads `output_range` / `model_channels` / `output_colorspace` from the processor rather than restating them in YAML, so the two cannot disagree. Plain `torch.nn.Module` criteria keep working untouched.

`bind` is abstract rather than defaulted for the same stated reason `SRProcessor.output_range` is: a wrong inherited default is exactly the failure the declaration prevents.

The one perceptual recipe reproducible here without a discriminator is Ledig et al.'s SRResNet-VGG22 — a VGG22 content loss plus total variation at `2e-8`, with **no pixel term** — which is why composition is in scope at all.

## Paper fidelity

Quoted rather than recalled: `φ_{i,j}` is the feature map after the j-th convolution *after activation* before the i-th maxpool (ESRGAN's pre-activation variant is a flag); feature maps are rescaled `1/12.75`, so an MSE of them carries `0.0061513` — the paper's "≈0.006"; total variation at `2e-8` applies to SRResNet-VGG22 **only**; and the MSE-initialisation trick is scoped to the GAN, so SRResNet-VGG22 trains from scratch.

Layer resolution is exercised against real torchvision: `conv2_2` → index 7 and `conv5_4` → 34 on VGG19, `conv5_3` → 28 on VGG16. Any `φ_{i,j}` works; a layer a depth does not have raises and names the deepest one that exists (`vgg54` is valid on VGG19, invalid on VGG16).

## Two silent-wrongness paths closed by construction

- **Range.** The same picture in `[0,1]` and in `[-1,1]` must score identically. A test asserts exactly that, and no-op'ing `bind` makes it fail — so the seam is verified, not asserted.
- **Colour space.** `SRProcessor` gained an abstract `output_colorspace`, so a VGG loss now refuses a 3-channel **non-RGB** processor (`YCbCrProcessor`) unless `allow_non_rgb: true`. Previously it was accepted silently: VGG received Y/Cb/Cr planes as R/G/B, normalised with RGB ImageNet statistics, and returned a healthy-looking loss for a run that reproduced nothing. The 1-channel case stays inside the pre-existing `grayscale_to_rgb` gate, so one flag covers one intent.

## Checkpoints stay small, and recipes stay interchangeable

The frozen truncated VGG is held outside the module tree (`object.__setattr__`, the existing `_compiled` precedent) with `_apply` overridden so `.to()` and dtype casts still reach it. A `.ckpt` therefore stays ~19 MB instead of ~99 MB, and an MSE-trained checkpoint still loads `strict=True` into a VGG-configured module. Only the slice actually evaluated is retained — vgg22 keeps 260,160 parameters, not 20,024,384.

Side effects worth knowing: the VGG cannot be flipped to `train()` by a parent, and a parent-wide `requires_grad_(True)` cannot unfreeze it — both strictly safer than a registered submodule. The cost is that it is invisible in `ModelSummary`.

Which loss trained a model is now recorded in checkpoint/ONNX provenance, including non-default knobs, since `before_activation` and `distance` change the objective materially.

## One bug worth reading about

The final review caught a Critical that no per-task review could see. `WeightedSumLoss` **rebound** `last_terms[name]` each forward. A CUDA-graph replay re-executes only recorded kernels, never that Python line, so only the tensor present at capture can ever be updated — and any interleaved *eager* forward (mid-training validation, or an epoch's partial last batch) repointed the entry at a tensor no replay would touch. Every `loss/train/<term>` tag would freeze at a validation batch's value while `loss/train` kept moving, and the "terms sum to the total" property would quietly stop holding.

That is the same failure class as #75 — mid-training validation severing a graph-owned binding. Per-term values are now written **in place** into stable buffers, which also made the contract CPU-testable: the guard asserts the same tensor object across forwards with a changed value, and it fails on the pre-fix code.

The test originally written to guard this **would have passed with the bug present** (four exact batches, no validation loop, so no eager step ever interleaved). It has been rewritten to compare graphed and eager per-term traces with a partial last batch forcing an interleave — and it is one of the seven still awaiting the GPU.

## Verification

- **612 passed / 9 skipped**; ruff clean; `pyproject.toml` unchanged — **no new dependency**, torchvision was already required.
- Every VGG in a test builds with `weights=None`, so the suite never touches the network. A real first use downloads ~548 MB once.
- The import-weight guards still hold: `sisr/cache.py` and `sisr/datasets/hr_cache.py` remain torch-free.
- The composite recipe is resolved **through the real CLI**, not constructed in Python, so the nested `class_path`/`init_args` path is a regression guard rather than an assumption.
- Reviews mutation-tested each silent mechanism — the grad asymmetry, the range mapping, the unregistered VGG, `_apply`, `bind` forwarding, and the weighted `last_terms` — confirming a test fails when each is broken.

## Known limitations

`ModelSummary` invisibility; per-term logging is a structural protocol rather than a declared member; a resume whose config omits `criterion:` silently reverts to MSE (a mismatch warning is a candidate follow-up); and step-time and peak-memory figures for VGG runs are still unmeasured, VGG54 at 96×96×b16 being the specific memory risk.

Docs ship separately in the stacked doc-only PR, so this branch stays code-only.

### DIFF
--- a/sisr/losses/__init__.py
+++ b/sisr/losses/__init__.py
@@ -1,6 +1,7 @@
 """Pluggable loss functions, wired via YAML ``model.criterion``."""
 
 from .base import SRLoss
+from .composite import WeightedSumLoss
 from .pixel import CharbonnierLoss, TotalVariationLoss
 from .vgg import VGG16FeatureLoss, VGG19FeatureLoss
 
@@ -10,4 +11,5 @@ __all__ = [
     "TotalVariationLoss",
     "VGG16FeatureLoss",
     "VGG19FeatureLoss",
+    "WeightedSumLoss",
 ]
--- a/sisr/losses/__init__.py
+++ b/sisr/losses/__init__.py
@@ -1,0 +1,5 @@
+"""Pluggable loss functions, wired via YAML ``model.criterion``."""
+
+from .base import SRLoss
+
+__all__ = ["SRLoss"]
--- a/sisr/losses/__init__.py
+++ b/sisr/losses/__init__.py
@@ -2,5 +2,12 @@
 
 from .base import SRLoss
 from .pixel import CharbonnierLoss, TotalVariationLoss
+from .vgg import VGG16FeatureLoss, VGG19FeatureLoss
 
-__all__ = ["SRLoss", "CharbonnierLoss", "TotalVariationLoss"]
+__all__ = [
+    "SRLoss",
+    "CharbonnierLoss",
+    "TotalVariationLoss",
+    "VGG16FeatureLoss",
+    "VGG19FeatureLoss",
+]
--- a/sisr/losses/__init__.py
+++ b/sisr/losses/__init__.py
@@ -1,5 +1,6 @@
 """Pluggable loss functions, wired via YAML ``model.criterion``."""
 
 from .base import SRLoss
+from .pixel import CharbonnierLoss, TotalVariationLoss
 
-__all__ = ["SRLoss"]
+__all__ = ["SRLoss", "CharbonnierLoss", "TotalVariationLoss"]
--- a/sisr/losses/base.py
+++ b/sisr/losses/base.py
@@ -15,6 +15,15 @@ class SRLoss(torch.nn.Module, abc.ABC):
     model. Plain :class:`torch.nn.Module` criteria (``MSELoss``, ``L1Loss``,
     :class:`~sisr.losses.pixel.CharbonnierLoss`) need no such adaptation and
     are used directly — this base exists only for the losses that do.
+
+    A loss may optionally expose ``last_terms: dict[str, torch.Tensor]``, a
+    structural protocol :class:`SRLightning` checks via ``getattr`` rather
+    than ``isinstance``, so any criterion holding one participates in
+    per-term logging as ``loss/<stage>/<name>``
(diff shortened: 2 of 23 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #77, opened 2026-08-05
Document the loss options and the SRResNet-VGG22 recipe

YeapJieShen:
Documents the loss options added in #76: a `## Losses` section in the README and commented recipes in the SRResNet template. No code.

**Stacked on #76** — merge that first. Base is `feat/losses`; retarget to `main` after #76 lands.

## Why this is a separate PR rather than a separate commit

This repo squash-merges, so one PR becomes exactly one commit. Pairing the docs with #76's code would flatten them into a single commit and break the standing rule that doc-only changes are their own commit. Splitting is what preserves it.

## What it says

The criterion is wired from YAML like everything else and defaults to `torch.nn.MSELoss`. Any `nn.Module` taking `(pred, target)` works, so **L1 needs no class of ours** — that is documented as a recipe rather than shipped as a wrapper. A table covers the six available losses, and the composite example is Ledig et al.'s SRResNet-VGG22: VGG22 at 1.0 plus total variation at `2e-8`, no pixel term, trained from scratch.

The template's default stays MSE and every recipe stays commented.

## The four traps it records

These are the things that cost real time to establish, so they are written down rather than left to be rediscovered:

- **First use downloads ~548 MB** of VGG19 weights into the torch hub cache. `weights: null` skips it but builds a random VGG and warns — tests only.
- **Total variation's `2e-8` presumes a `[-1, 1]` output range** (`RGBSignedOutputProcessor`, which the template already wires). Under a `[0, 1]` processor the same image has exactly half the total variation, so the effective weight differs by 2×.
- **A VGG loss refuses non-RGB input.** 1-channel needs `grayscale_to_rgb: true`; 3-channel non-RGB (`YCbCrProcessor`) needs `allow_non_rgb: true`. The reason is that VGG normalises with RGB ImageNet statistics, so Y/Cb/Cr planes fed as R/G/B produce a healthy-looking loss over meaningless features.
- **A perceptual run scores worse PSNR by design**, so the PSNR-monitored "best" checkpoint stops tracking the training objective. The template already writes SSIM-monitored checkpoints alongside, so this is a choice of which artifact you want rather than setup work.

Layer names are given as per-block convolution counts rather than a range, because `vgg11`–`vgg54` reads contiguous while `vgg13`, `vgg23` and `vgg35` do not exist. VGG16 is marked as offered for experimentation rather than fidelity — no published SR number uses it.

## Verification

Both the shipped template and a copy with the recipe uncommented resolve through the real CLI (`--print_config`), the latter yielding `sisr.losses.WeightedSumLoss` with both terms — so the commented YAML is valid once uncommented, not merely plausible. Every factual claim was checked against the code on this branch, including the ~548 MB figure (574,673,361 bytes). ruff clean; no `.py` file touched.

Still missing, pending measurement: step-time and peak-memory figures for VGG runs, so a first `vgg54` launch has no documented cost expectation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,65 @@ means writing a network and a config dataclass, not a new Lightning module. See
 | SRCNN | [Dong et al. 2015](https://arxiv.org/pdf/1501.00092) | pre-upsampled to HR size | none, same-resolution refinement |
 | SRResNet | [Ledig et al. 2017](https://arxiv.org/pdf/1609.04802) | true low-resolution | ×scale sub-pixel convolution |
 
+## Losses
+
+The criterion is wired from YAML like everything else, and defaults to
+`torch.nn.MSELoss`. Any `nn.Module` taking `(pred, target)` works, so L1 needs no
+class of ours:
+
+```yaml
+model:
+  criterion:
+    class_path: torch.nn.L1Loss
+```
+
+| Class | What it is |
+|---|---|
+| `torch.nn.MSELoss` | the default; both papers' baseline |
+| `torch.nn.L1Loss` | plain L1 |
+| `sisr.losses.CharbonnierLoss` | `sqrt(diff² + eps²)` — L1 with a finite gradient at zero |
+| `sisr.losses.TotalVariationLoss` | isotropic TV regulariser; ignores its target |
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #79, opened 2026-08-06
Compare per-term loss traces to a tolerance, not exactly

YeapJieShen:
Fixes a flaky test I introduced in #76. It went green seven consecutive times on the branch and in CI, then failed on the first full-suite run against `main` — which is the worst way for a test to be wrong.

```
FAILED tests/training/test_integration.py::test_cuda_graph_replay_keeps_per_term_losses_live
At index 11: tv 2.7041083740186878e-05 != 2.7041087378165685e-05
1 failed, 620 passed
```

## What was wrong

The test compares a graphed run's per-term loss trace against an eager run's, and demanded **bit-equality**. Its sibling `test_cuda_graph_loss_trace_is_bit_identical_to_eager` can legitimately demand that — it compares a single MSE total. The `tv` term cannot: it reaches float32 through a `sqrt`, and these tests run with `benchmark=True`, so cuDNN autotunes and can select different algorithms per process. The two runs then differ in the last ulp.

The observed gap is **~1.3e-7 relative**, which is float32 epsilon. Nothing was actually wrong with the code under test.

## The fix

Per-entry comparison at `rel=1e-5`, with the reasoning recorded in the docstring so nobody tightens it back later.

That tolerance is deliberately placed between two measured quantities rather than picked by feel:

| | relative size | |
|---|---|---|
| float32 ulp noise (observed) | 1.3e-7 | ~80× below the tolerance |
| `rel=1e-5` | 1e-5 | |
| the bug this guards (measured under mutation) | 3.9e-2 | ~4000× above the tolerance |

## Verification

- **Stability:** 5 consecutive runs, 5 passed / 0 failed.
- **Still catches the bug:** restoring the pre-fix rebind in `WeightedSumLoss.forward` makes it fail at **step 4** — the first partial-batch eager fallback, exactly the interleave the test is built to create:
  ```
  AssertionError: (4, 'mse')
  assert 0.3428382873535156 == 0.3298541307449341 ± 3.3e-06
  ```
  Mutation reverted; working tree clean.
- **Full suite on this branch, GPU visible and `data/` present: 621 passed, 0 skipped, 0 failed.** ruff clean.

Test-only change; no source file is touched.

### DIFF
--- a/tests/training/test_integration.py
+++ b/tests/training/test_integration.py
@@ -540,7 +540,7 @@ def test_fast_dev_run_logs_per_term_loss_tags_for_a_composite_criterion(
 
 
 class _TermTrace(lightning.Callback):
-    """Records the per-step ``last_terms`` values, for exact cross-run comparison."""
+    """Records the per-step ``last_terms`` values, for cross-run comparison."""
 
     def __init__(self):
         self.terms: list[dict[str, float]] = []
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #80, opened 2026-08-07
Add daala-methodology SSIM as a selectable evaluation convention

YeapJieShen:
## Why

Ledig et al. (SRResNet/SRGAN) state their published PSNR/SSIM were computed *"on the y-channel of center-cropped, removal of a 4-pixel wide strip from each border, images using the **daala** package"*.

daala's SSIM is not Wang's. Its gaussian sigma **scales with frame height** (`_h*(1.5/256)`), where Wang's — what torchmetrics, MATLAB and BasicSR compute, and therefore what most SR papers report — is fixed at 11×11, σ=1.5.

That single difference accounted for the project's entire SSIM gap against the paper, while PSNR already matched. A **training-free control** established it before any code was written: scoring plain bicubic ×4 through this repo's own stack reproduced the paper's bicubic *PSNR* while missing its bicubic *SSIM* by −0.0105 / −0.0432 / −0.0238 — the same per-dataset offsets the trained model showed, with no model involved.

## What

- **`sisr/ssim.py`** — a port of daala's methodology: integer kernel summing to exactly 256 (centre tap absorbs the rounding), truncate-and-renormalise borders, weight-weighted pooling, 8-bit input, float64 throughout.
- **`SREvalConfig.ssim_impl: Literal["wang", "daala"] = "wang"`** — validated at construction. Switching is **in place**: the `ssim/...` tag names are unchanged, so a figure is comparable only to one computed under the same setting. Recorded in `hparams` and every artifact's `sisr_meta`.
- **`SRLightning._mean_ssim`** — one dispatcher. Previously the validation path and the benchmark callback each called torchmetrics directly; either could have honoured the flag while the other silently didn't.
- **`SRResNetEvalConfig.ssim_impl = "daala"`** — the architecture default, same precedent as `crop_border = 4`. **SRCNN and the base config stay `"wang"`.**

## Evidence

**`tests/reference/daala_ssim.c`** vendors daala's own `gaussian_filter_init` + `calc_ssim` **verbatim** (BSD-2-Clause, Xiph.Org) as a ground-truth oracle. Expected values are **committed**, so CI needs no C toolchain and nothing skips.

- **133/133 parity cases at `rel=1e-9`** — 14 synthetic (always run) + 119 real Set5/Set14/BSD100 (data-gated). Worst relative difference **3.324e-14**.
- **CUDA path proven.** float64 conv on CUDA could have dispatched to an FFT-based cuDNN algorithm, which would break the exact-integer accumulation the whole port rests on. Measured: 9/14 bit-identical, 5/14 within one float64 ulp (≤3.16e-16). Not FFT.
- **Validation-path cost: none measurable.** 108.863 s/epoch (wang) vs 107.987 s (daala), 219 images. n=1 per config, so recorded as a bound that rules out any large effect, not a point estimate.

Re-scoring the finished 1M-step golden checkpoint moves SSIM from −0.0103 / −0.0385 / −0.0274 against the paper to **−0.0017 / −0.0017 / −0.0032**, with PSNR reproducing to ~1e-4 dB as a control. The residual is now uniform across the three sets instead of varying 4× — that variation was the convention.

## Notes for review

- **The tolerance is `rel=1e-9`, deliberately, and should not be tightened.** The pooling divergence between the C's sequential row-major accumulation and torch's tree reduction grows as ≈`eps·n/4`, so a *correct* implementation exceeds `1e-12` at 256² on flat content and reaches 1.5e-11 at 1024². Measured defect signals are ≥4.96e-8. `1e-9` sits in the empty band between them.
- **`tests/reference/daala_ssim.c` is vendored third-party code, kept byte-for-byte.** Its formatting and unused locals are upstream's. `tests/test_packaging.py` pins `packages = ["sisr"]` so it can never ship to users.
- `pythonpath = ["."]` was added to the pytest config: this branch introduces the repo's first `from tests.…` import, which otherwise resolved only via the editable-install `.pth` and would silently mix trees under the worktree workflow.
- `test_ckpt_path_loses_subclass_only_eval_defaults` documents a pre-existing trap this work surfaced — see the docs PR.

Docs follow in a separate PR (this repo squash-merges, so pairing them would flatten the doc-only-separate-commit rule).


YeapJieShen (commented):
Human reviewed, all good.

PR #81, opened 2026-08-07
Document the two SSIM conventions and the ssim_impl override

YeapJieShen:
⛔ **Stacked on #80 — do not merge until #80 lands.** Base retargets to `main` automatically when #80 merges. CI does not run while the base is a feature branch; it fires on the rebase+push that follows #80's merge.

Doc-only counterpart to #80. Separate PR rather than a separate commit because this repo squash-merges — pairing them would flatten the standing doc-only-separate-commit rule.

## What

**`README.md` → Comparability** — explains why SSIM has two incompatible conventions and PSNR does not: PSNR is a closed-form function of squared error, so any correct implementation agrees; SSIM depends on a local-window gaussian the field never standardised. Wang's is fixed 11×11 σ=1.5; daala's sigma scales with image height. States the hazard plainly — an SRResNet SSIM figure is comparable to Ledig et al. and **not** to Wang-based tables (EDSR/RCAN/SwinIR/BasicSR lineage) — and notes that because the switch is in place, neither the metric tag nor the checkpoint filename reveals which convention produced a number.

**Both templates** — comments on the bare `eval_config:` blocks explaining which default applies and how to override, in each architecture's own terms.

**`README.md` → a new section on `--ckpt_path`.** This documents a **pre-existing trap** that this work surfaced rather than caused:

`SRLightningCLI._parse_ckpt_path` reconstructs `eval_config` from a checkpoint's stored hparams as a **bare dict with no `class_path`**, so jsonargparse instantiates the *annotation* type — base `SREvalConfig` — **discarding the subclass and every subclass-only default**. Fields the checkpoint stores explicitly (`crop_border`, `psnr_channels`) survive; fields it doesn't fall back to the base default.

Consequences worth knowing:
- Re-scoring a pre-change checkpoint silently uses `wang`, and a command-line `--model.eval_config.ssim_impl=daala` does **not** help.
- It applies to `fit --ckpt_path` resume too, which is the more damaging case: a resumed run trains and *checkpoints* on `wang` `ssim/val/RGB` while a fresh `fit` from the identical YAML uses `daala`.
- **It is general** — any future `SREvalConfig` field with an architecture override will behave this way. `test_ckpt_path_loses_subclass_only_eval_defaults` (in #80) is the executable statement of the rule.

The workaround snippet handles both the nested and the legacy `/`-flattened hparams formats, and was verified by execution against a real checkpoint (original confirmed unmodified by SHA-256).

No paper-vs-ours results table — that publication decision remains deferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,35 +169,79 @@ both are pinned:
   [`SRResNetEvalConfig`](sisr/models/srresnet/config.py) overrides it to `'daala'`
   because that is the convention its paper used; SRCNN keeps `'wang'`, the field
   standard. The switch is **in place**: `ssim/val/RGB` and `ssim/val/Y` name the
-  metric identically either way, so the convention is not visible in the tag — it is
-  recorded in `hparams` and in every artifact's `sisr_meta` instead. Consequently, an
+  metric identically either way, and checkpoint filenames
+  (`sr-{step}-ssim_val_RGB={value:.4f}.ckpt`, built by
+  [`SRCheckpoint`](sisr/training/callbacks.py)) carry only the bare number — so
+  neither the tag nor the filename reveals which convention produced a given value.
+  It is recorded in `hparams` and in every artifact's `sisr_meta` instead. Consequently, an
   SRResNet SSIM figure is comparable to Ledig et al. and **not** to Wang-based tables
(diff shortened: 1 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #83, opened 2026-08-08
docs: document the delivery workflow, commit conventions, and security policy

YeapJieShen:
Adds three documents. No code changes, so `squash` is the right merge here.

## Why

The delivery workflow and the evidence expectations were real but undocumented — they lived in review comments and habit, so a contributor could only learn them by getting a PR corrected.

## What changed

**`CONTRIBUTING.md`** — a "Making a change" lifecycle table (design → branch → verify → PR → merge), an **Evidence** section, and a rewritten **Commits** section.

Two rules are worth calling out because they interact:

1. **Code and docs may share a PR but never a commit.**
2. **Conventional Commit types** (`feat`/`fix`/`docs`/`perf`/`refactor`/`test`/`chore`).

The second exists to make the first checkable. "Separate commits" is unverifiable by anything without a marker in the subject; `docs:` makes the split visible in `git log`.

This also means **merge strategy now matters per PR**: a squash would collapse a `docs:` commit back into its code commit and silently defeat rule 1. So single-purpose PRs squash, and mixed code+docs PRs **rebase-merge** — which also satisfies the linear-history requirement.

The subject convention deliberately preserves what this repo already does well: subjects describe the **effect**, not the mechanism (`fix: make the cache build lock unable to destroy live data`, not `fix: refactor _try_load`).

**`SECURITY.md`** — states a contract the code already keeps. Every checkpoint read in `sisr/` passes `weights_only=True` (`export.py`, `cli.py`), and `training/metadata.py` is shaped around staying loadable under it — plain types only, `str()`-coerced `torch.__version__`, configs stored via `asdict`. That was previously recorded only in a `pyproject.toml` comment. Documenting it means a change that breaks it reads as a security regression rather than an inconvenience.

It is also explicit about what the project does **not** guarantee: `weights_only=True` narrows the attack surface but is not a sandbox, Lightning's full-checkpoint resume path is outside this project's control, and a `class_path` config is executable by design.

**`CHANGELOG.md`** — Keep a Changelog format, everything under `Unreleased`, no tags. The project is pre-1.0 with no releases, so tagging now would imply a stability commitment that does not exist.

## Verification

Docs-only; no tests affected. Content was checked against the code rather than transcribed from memory — the `weights_only=True` call sites and the `torch>=2.6` CVE floor were both verified in-tree before being described.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,74 @@
+# Changelog
+
+Notable changes to this project, following [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+Versioning will follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once a
+release is tagged. Nothing is released yet: the project is pre-1.0 and every entry below
+sits under **Unreleased**. Until a version is cut, any part of the API may change without
+notice.
+
+## [Unreleased]
+
+### Added
+
+- **SRResNet** and **SRCNN**, both fully wired — model, dataset pipeline, config
+  dataclasses, and an experiment template each.
+- **`sisr` console script** over a single `LightningCLI` entrypoint, with `fit`, `test`,
+  `predict` and `export` subcommands. One self-contained YAML drives an experiment.
+- **Pluggable losses** (`sisr/losses/`): `CharbonnierLoss`, `TotalVariationLoss`,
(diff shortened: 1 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #86, opened 2026-08-08
feat: run LightningCLI in subclass mode so a config can name its Lightning module

YeapJieShen:
⛔ **Do not merge yet — bottom of a 6-PR stack.** Merge order is strict; see *Stack* below.

Runs `LightningCLI` with `subclass_mode_model=True`, so `model:` in a config carries `class_path` + `init_args` like every other component and a YAML can name its Lightning module.

## Why

`model_class=SRLightning` was fixed at the CLI, so **no config could select a different Lightning module** — the blocker for the adversarial training module later in this stack. `model:` was also the sole component in the config not following the project's own `class_path` + `init_args` convention, so this removes an inconsistency rather than adding one.

## What changed

- `sisr/cli.py` — `subclass_mode_model=True`; optimizer/lr_scheduler links retargeted to `model.init_args.*`.
- Both templates re-nested under `class_path: sisr.training.SRLightning` / `init_args:`. Every existing comment was preserved verbatim and moved with the key it documents — they record measured findings (the VGG22 + total-variation recipe, the SSIM convention default, why `example_input_shape` exists).
- `sisr/cli.py` — `_parse_ckpt_path` now merges the checkpoint's hyperparameters through the **subcommand's own parser**, not the top-level one.

## The `_parse_ckpt_path` fix, precisely

The obvious one-line change (nesting the injected dict one level deeper while still parsing through `self.parser`) **does not work**, and was measured failing before the real fix was written.

The cause is `jsonargparse`'s `ActionTypeHint._check_type`, which does `prev_val = cfg.get(self.dest)` where `self.dest` is the action's **parser-local** name (`"model"`). That is correct when `cfg` is already scoped to the subcommand's parser, and a **silent miss — not an error** — when reached through the top-level dispatching parser, where the value actually lives one level deeper at `cfg["<subcommand>"]["model"]`.

Routing through `self._parser(subcommand)` fixes it. The docstring records the mechanism and carries an explicit warning against "simplifying" it back, because the failure is silent in exactly the fields that have a usable fallback.

## Side effect: checkpoint reload now preserves eval-config subclass identity

Reloading via `--ckpt_path` previously rebuilt `eval_config` as the **base** class, silently dropping every default an architecture subclass had overridden — so a resumed SRResNet run could monitor a different SSIM convention than a fresh run from the identical YAML, distinguishable only by reading `hparams`.

That shares its root cause with the parser-scoping bug above and is fixed by the same change. The characterisation test that documented the defect has been **replaced, not deleted**, per its own recorded acceptance criteria: `test_ckpt_path_preserves_eval_config_subclass_identity` now asserts two independently-failable things — that the reloaded object is the subclass, and that a subclass-only default (`ssim_impl == "daala"`) survives from a checkpoint whose stored dict **never contained that key**.

**Evidence caveat, stated plainly:** a clean RED for this specific guarantee is structurally unreachable in this configuration. Under subclass mode `model` is Union-typed and required with no fallback, so any un-nested state crashes there before `eval_config` resolution is ever reached — the pre-fix failure is a construction-time `SystemExit`, not the silent reversion itself. The forward guarantee is directly asserted; the RED half is a generic failure. This limitation is recorded in a comment on the test rather than only in review notes.

## Testing

796 passed / 0 skipped / 0 failed. `ruff check` and `ruff format --check` clean. No skips — the two data-gated byte-equality tests have their data on the dev box and ran.

## Stack

```
main
 └─ THIS PR  · CLI subclass mode
         └─ rolling checkpoints
                     └─ docs
```

Merge this one first; GitHub retargets the next PR's base to `main` automatically. **Do not use "delete branch on merge"** — deleting the head branch races and closes the child PR instead of retargeting it.

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -267,8 +267,12 @@ class SRLightningCLI(LightningCLI):
         parsed through the *subcommand's own* parser/config branch
         (``self._parser(subcommand)``, ``self.config[subcommand]``), never through
         the top-level subcommand-dispatching one (``self.parser``, ``self.config``),
-        even though both expose the same ``model.init_args.<key>`` option path and
-        neither raises at parse time.
+        even though both expose the same ``model.init_args.<key>`` option path.
+        The top-level route parses without a syntax error, but per-field the
+        outcome is either a hard ``SystemExit`` (``model``, ``processor``: no
+        default, required) or a silent revert to the bare-annotation default
+        (any nested subclass field with a fallback, e.g. ``eval_config``) — see
+        below for which and why.
 
         The reason: jsonargparse's subclass-merge machinery
         (``ActionTypeHint._check_type``) looks up the field's *previous* value as
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -223,13 +223,14 @@ class SRLightningCLI(LightningCLI):
     def add_arguments_to_parser(self, parser):
         """Wire top-level ``optimizer:`` / ``lr_scheduler:`` / ``matmul_precision:`` keys.
 
-        Non-subclass mode (``model_class=SRLightning`` fixed) means
-        ``SRLightning``'s init args land at ``model.<arg>``, not
-        ``model.init_args.<arg>`` — that's why the link targets omit
-        ``init_args``.
+        Subclass mode (``subclass_mode_model=True``) means the module's init args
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #87, opened 2026-08-08
feat: add LPIPS and DISTS perceptual evaluation metrics

YeapJieShen:
⛔ **Do not merge yet — 2nd of a 6-PR stack.** Based on `feat/cli-subclass-mode`; merge that one first.

Adds LPIPS and DISTS as evaluation metrics, reported for the validation set and for each benchmark set.

## Why

An adversarial objective makes PSNR and SSIM **worse by design**, so the SRGAN work later in this stack would otherwise ship a training mode with no metric that tracks what it is optimising. LPIPS and DISTS correlate with human judgement better than PSNR does. Neither is a substitute for the paper's MOS study, and nothing here claims otherwise.

## What changed

- **`sisr/perceptual.py`** — `PERCEPTUAL_METRICS` (name → is-lower-better) and `perceptual_score(...)`. The one place either library's argument conventions are known.
- **`SREvalConfig`** — `perceptual_metrics: list[str] = []` and `lpips_net: Literal["alex","vgg","squeeze"] = "alex"`, plus a `perceptual_keys` property mirroring `psnr_keys`/`ssim_keys`.
- **`SRLightning._mean_perceptual`** — the single decision point for which perceptual metric this project computes, beside `_mean_ssim` for the same reason that one exists.
- **Tags** — `lpips/val` · `dists/val` · `lpips/{set}` · `dists/{set}`. No colorspace segment: both metrics are RGB-only by definition.
- **`_validate_monitor_metric`** — now also validates metric *direction*.
- **Packaging** — new `perceptual` extra (`lpips`); `torchmetrics` floor raised to `>=1.9`.

## Defaults are off, deliberately

`perceptual_metrics` is empty by default, so **every existing SRCNN/SRResNet run logs exactly the tags it logged before**. A test asserts no perceptual tag appears when unrequested.

## Two argument conventions, both silent-failure shaped

These are why the module exists as a seam rather than two inline calls:

1. **LPIPS needs `normalize=True`.** Its default `False` means "inputs are already `[-1,1]`"; this project's tensors are RGB `[0,1]`, so the default silently scores a squashed range — a plausible number, no error.
2. **The two disagree on reduction defaults** (LPIPS `'mean'`, DISTS `None`), so one would return a scalar and the other a per-image tensor under the same logging call. Both are passed explicitly.

Both are pinned by tests whose spies declare the parameters **without defaults**, so omitting either fails with `TypeError` rather than silently passing a wrong value.

## Monitor direction is now validated

Both checkpoint classes default to `mode="max"` — correct for PSNR/SSIM, exactly inverted for LPIPS/DISTS. `monitor_metric: lpips/val` at the default would have kept the **worst** checkpoint of the run, with nothing in the logs, tags or filenames revealing it. The validator now rejects the mismatch and names the mode to set instead. Direction is looked up per metric from `PERCEPTUAL_METRICS`, not restated.

All eight checkpoint entries across the three shipped templates monitor `psnr/val/RGB` or `ssim/val/RGB` at the default mode and continue to validate unchanged.

## Dependencies

**DISTS costs nothing** — torchmetrics implements it over torchvision's VGG16, both already required. **LPIPS needs the `lpips` package**, so it sits behind `pip install '.[perceptual]'` with an error naming the extra. The `torchmetrics>=1.9` floor is the version DISTS was *verified* present in here; it was not lowered to a version nobody tested.

## Testing

813 passed / 0 skipped / 0 failed. `ruff check` and `ruff format --check` clean.

**CI stays hermetic:** the torch hub cache is empty and CI has no network, so every test monkeypatches at a seam that is actually reached — none can trigger a weights download (~230 MB AlexNet / ~528 MB VGG16). Each new test ships with the mutation that proves it fails; one test that mocked a constant, and so was blind to the tensor choice it was named for, was caught in review and rebuilt to depend on its inputs.

## Stack

```
main
 └─ CLI subclass mode
     └─ THIS PR  · perceptual metrics (LPIPS/DISTS)
         └─ rolling checkpoints
                     └─ docs
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,10 @@ dependencies = [
     "torch>=2.6",
     "torchvision>=0.21",
     "lightning>=2.4",
-    "torchmetrics>=1.4",
+    # 1.9 is the version DISTS was verified present in on this project
+    # (torchmetrics.functional.image.deep_image_structure_and_texture_similarity).
+    # Lowering it means testing the lower version, not guessing.
+    "torchmetrics>=1.9",
     "jsonargparse[signatures]>=4.27",
     "lmdb>=1.4",
     "pillow>=10",
@@ -72,6 +75,11 @@ export = [
     "onnx>=1.16",
     "onnxruntime>=1.19",
 ]
+# LPIPS only. torchmetrics implements both LPIPS and DISTS, but gates LPIPS on
+# the `lpips` package (_LPIPS_AVAILABLE) while DISTS needs only torchvision,
+# already a core dep — so DISTS ships in core and this extra buys exactly one
+# metric. Optional because it is a scoring convention, not a training need.
+perceptual = ["lpips>=0.1"]
 
 [project.scripts]
 sisr = "sisr.cli:main"
--- a/sisr/perceptual.py
+++ b/sisr/perceptual.py
@@ -1,0 +1,85 @@
+"""LPIPS / DISTS perceptual metrics — the two conventions, in one place.
+
+Both are *scoring* metrics, never losses: LPIPS postdates Ledig et al. by a
+year, so an LPIPS-supervised generator is a different method, not SRGAN.
+
+They exist here because an adversarial objective makes PSNR and SSIM worse by
+design, so without a perceptual metric an SRGAN run has no quality signal at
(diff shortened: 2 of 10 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #88, opened 2026-08-08
feat: add rolling last-N checkpoints and component weight files

YeapJieShen:
⛔ **Do not merge yet — 3rd of a 6-PR stack.** Based on `feat/perceptual-metrics`; merge the two below it first.

Adds a rolling last-N checkpoint mode, and lets one checkpoint callback save **any** named component's bare weights with provenance that describes *that component*.

## Why rolling mode

Under an adversarial objective PSNR and SSIM get **worse by design**. A metric-monitored `save_top_k` therefore selects the **least adversarial** model — very likely a checkpoint from the first few thousand steps. For a GAN, metric-monitored selection is not merely uninformative, it is inverted. Rolling mode keeps a window of recent checkpoints instead.

## How it had to be built

Lightning **refuses** `ModelCheckpoint(monitor=None, save_top_k=3)` outright:

```
MisconfigurationException: ModelCheckpoint(save_top_k=3, monitor=None) is not a
valid configuration. No quantity for top_k to track.
```

So a rolling window cannot be expressed through `save_top_k` at all. The working construction is `monitor=None, save_top_k=-1` (keep everything) with the window enforced by our own oldest-first deletion through `ModelCheckpoint._remove_checkpoint`. The filename must carry `{step}` or every save overwrites the last.

**One structural trap worth recording:** the deletion hook cannot live in a mixin that overrides `_save_checkpoint`. `SRWeightsCheckpoint` already overrides that method to write bare `.pt` weights and never calls `super()`, and Python resolves the class's own attribute before any base — so a mixin's version would have been **shadowed and silently never run**, leaving `.pt` files to accumulate forever with nothing erroring. The logic is therefore a separate `_enforce_rolling_window` seam called explicitly from each class's `_save_checkpoint`, with a comment at both call sites saying the call must be preserved.

## Component weights and matching provenance

`SRWeightsCheckpoint` gains `attribute: str = "model"`, so one callback can emit both `sr-weights-*.pt` (generator) and `d-weights-*.pt` (a second network). Their distinct `filename_prefix` and `FILE_EXTENSION` already keep each one's deletion pass off the other's files.

A non-generator file must **not** carry `build_metadata`'s output — that describes the generator (`io.scale`, `io.input`, `criterion`, `eval_config`), and attaching it to another network's weights would describe something the file does not contain. That is precisely the silent-wrong-artifact class this metadata exists to prevent. So `metadata.py` gains a second builder:

```
build_component_metadata(module, attribute) ->
  {format, kind: "component", created, versions,
   component: {name, class_path, init_args},
   io: {input_range, input_colorspace},
   training: {...}}
```

`io.input_range` is the load-bearing field: a discriminator is trained on model-space `[-1, 1]` tensors, and feeding it `[0, 1]` later is wrong with no error.

Both builders share one private `_envelope(...)` — the module's whole premise is one function feeding several sinks so the payloads cannot drift. That also preserves the existing `str(torch.__version__)` coercion, without which `torch.load(weights_only=True)` refuses the file (`TorchVersion` is a `str` subclass it won't unpickle).

**`FORMAT_VERSION` is not bumped** — this is a new artifact *kind*, not a change to an existing one. A `kind` field is added to both builders; its absence on pre-existing files means `"sr_model"`, and the module docstring says so.

## Testing

830 passed / 0 skipped / 0 failed. `ruff check` and `ruff format --check` clean.

Every new test ships with the mutation that proves it fails. One mutation — removing the `monitor is not None` guard — **slipped past every originally-planned test**, which is what prompted the extra test that now covers it.

A narrowly-scoped `filterwarnings` ignore is added for Lightning's manual-optimization checkpoint warning ("the checkpoint will contain the model state AFTER optimization"). That state is exactly what a checkpoint should hold, but this repo runs `filterwarnings = ["error", ...]`. It is unexercised on this branch and lands here for the adversarial work above it.

## Stack

```
main
 └─ CLI subclass mode
         └─ THIS PR  · rolling checkpoints + component weights
                     └─ docs
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,10 @@ filterwarnings = [
     # stops calling torch's deprecated pytree API (track the next lightning/torch bump).
     # Upstream: https://github.com/Lightning-AI/pytorch-lightning (pytree LeafSpec deprecation)
     "ignore::FutureWarning:lightning.*",
+    # Manual optimization (SRGANLightning) + every_n_train_steps: Lightning warns
+    # that the saved state is post-optimization. That is exactly the state we
+    # want — a checkpoint should hold the weights the step produced.
+    "ignore:Using ModelCheckpoint with manual optimization:UserWarning",
 ]
 
 [tool.ruff]
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -25,7 +25,7 @@ from lightning.pytorch.callbacks import BasePredictionWriter, Callback, ModelChe
 from lightning.pytorch.utilities.exceptions import MisconfigurationException
 
 from ..perceptual import PERCEPTUAL_METRICS
-from .metadata import build_metadata
+from .metadata import build_component_metadata, build_metadata
 
 
 class BenchmarkSample(NamedTuple):
@@ -768,6 +768,8 @@ class SRCheckpoint(_RollingSaveMixin, ModelCheckpoint):
             filepath: Destination path, already formatted by ``ModelCheckpoint``.
         """
         super()._save_checkpoint(trainer, filepath)
+        # Any edit here must preserve this call -- see _RollingSaveMixin's
+        # docstring for why a rolling window can't be expressed via save_top_k alone.
         self._enforce_rolling_window(trainer, filepath)
 
     def setup(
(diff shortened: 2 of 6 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #89, opened 2026-08-08
feat: add the SRGAN discriminator, adversarial loss, and configs

YeapJieShen:
⛔ **Do not merge yet — 4th of a 6-PR stack.** Based on `feat/rolling-checkpoints`; merge the three below it first.

The SRGAN components: discriminator, adversarial objective, and paper-faithful configs. **No training loop yet** — that is the next PR, so nothing here changes any existing run's behaviour.

Reference: Ledig et al., [*Photo-Realistic Single Image Super-Resolution Using a Generative Adversarial Network*](https://arxiv.org/pdf/1609.04802).

## `SRDiscriminator` — Ledig Figure 4

`k3n64s1 + LeakyReLU` → 7 × `(Conv-BN-LeakyReLU)` at `(64 s2, 128 s1, 128 s2, 256 s1, 256 s2, 512 s1, 512 s2)` → `Dense(dense_features)` → LeakyReLU → `Dense(1)`.

Three choices that are deliberate and documented in the code, because each has an obvious-looking "fix" that would be wrong:

- **Not an `SRModel`.** It is a classifier, not an SR mapping, so `input_contract` and `reset_parameters` would be contracts it cannot honour. It is a plain `nn.Module` carrying `hparams` purely so provenance metadata can describe it.
- **Emits logits — no sigmoid.** Figure 4 ends in a sigmoid; pairing an unactivated output with `BCEWithLogitsLoss` computes the identical objective, numerically stably. Restoring the sigmoid would **double-activate**.
- **`hr_input_size` is declared, not inferred**, because the dense head fixes the accepted input size. `Linear.in_features` derives from it (four stride-2 convs → `s = hr_input_size // 16`). A later PR validates the declared value against the real training crop, which is why it must be explicit. `dense_features` is a real argument whose default is the paper's 1024.

## `AdversarialLoss`

Non-saturating generator term plus the two-term discriminator objective, over logits.

**Deliberately not an `SRLoss`:** that ABC's contract is `forward(pred, target)` in model space, and the generator's term `−log D(G(I_LR))` **has no target**. Bending it to fit would put a dummy argument into the one signature the whole loss library depends on. Two named methods instead — which also makes the objective testable without a training loop and swappable from YAML later (LSGAN, relativistic) without touching the training module. Parameter-free, so it adds nothing to any `state_dict`.

`discriminator_loss` expects **already-detached** fake logits; it does not detach internally, which would silently mask a caller's mistake.

## Configs

`SRGANTrainingConfig(SRResNetTrainingConfig)` — the generator *is* SRResNet, so `scale=4` and the rest are inherited. Adds `init_from` (optional; unset trains from scratch), `adversarial_weight=1e-3` (the paper's), and `d_steps_per_g_step=1` — Goodfellow's `k`, where 1 is Ledig's 1:1 alternation.

`cuda_graph=True` is **refused**: the graph machinery captures one `{zero_grad, forward, loss, backward}` around **one** optimizer, and this mode will run two alternately under manual optimization. `super().__post_init__()` runs first so the inherited checks still apply.

`SRGANEvalConfig(SRResNetEvalConfig)` inherits `crop_border=4`, `psnr_channels=['RGB','Y']` and `ssim_impl='daala'`, so SRGAN numbers stay comparable to the SRResNet baseline, and turns perceptual metrics on (`['lpips','dists']`) because PSNR and SSIM get *worse* by design here.

That last one is a subclass-only **default**, not a subclass-only **field** — the fields live on base `SREvalConfig`. The distinction is load-bearing: a subclass-only field is dumped into a checkpoint by `dataclasses.asdict` and then handed to the base class on reload as an unexpected keyword argument, which is a hard failure.

## Testing

851 passed / 0 skipped / 0 failed. `ruff check` and `ruff format --check` clean.

Every test ships with the mutation that proves it fails. One result worth recording: the generator-loss **gradient-direction** test cannot distinguish the non-saturating form from the saturating one — both produce a strictly negative gradient at every input. Only the test pinning the formula against an independently-computed `binary_cross_entropy_with_logits` value catches that swap. A test that merely checked the loss was finite and pushed logits the right way would have been decorative.

Also folded in: the discriminator's size-validation message now says "positive multiple" (it rejects `0` and negative multiples of 16, for which "divisible by 16" was misleading) and interpolates the computed stride factor instead of hardcoding `16`, which would have drifted if the block table changed.

## Stack

```
main
 └─ CLI subclass mode
         └─ rolling checkpoints + component weights
             └─ THIS PR  · SRGAN discriminator, adversarial loss, configs
                     └─ docs
```

### DIFF
--- a/sisr/losses/__init__.py
+++ b/sisr/losses/__init__.py
@@ -1,5 +1,6 @@
 """Pluggable loss functions, wired via YAML ``model.criterion``."""
 
+from .adversarial import AdversarialLoss
 from .base import SRLoss
 from .composite import WeightedSumLoss
 from .pixel import CharbonnierLoss, TotalVariationLoss
@@ -7,6 +8,7 @@ from .vgg import VGG16FeatureLoss, VGG19FeatureLoss
 
 __all__ = [
     "SRLoss",
+    "AdversarialLoss",
     "CharbonnierLoss",
     "TotalVariationLoss",
     "VGG16FeatureLoss",
--- a/sisr/losses/adversarial.py
+++ b/sisr/losses/adversarial.py
@@ -1,0 +1,56 @@
+"""The SRGAN adversarial objective (Ledig et al. §2.2)."""
+
+import torch
+
+
+class AdversarialLoss(torch.nn.Module):
+    """Non-saturating GAN objective over discriminator **logits**.
+
+    Deliberately **not** an :class:`~sisr.losses.base.SRLoss`: that contract is
+    ``forward(pred, target)`` in model space, and the generator's term
+    ``-log D(G(I_LR))`` has no target at all. Bending it to fit would put a
+    dummy argument in the signature the whole loss library depends on. Being a
+    separate class also makes the objective testable without a training loop
+    and swappable from YAML (LSGAN, relativistic) without touching the module.
+
+    Takes logits, not probabilities — see
+    :class:`~sisr.models.srgan.SRDiscriminator`. Parameter-free, so it adds
+    nothing to any ``state_dict``.
+    """
+
+    def generator_loss(self, logits_fake: torch.Tensor) -> torch.Tensor:
+        """Generator's adversarial term: make the discriminator call fakes real.
+
+        The **non-saturating** form (``-log D(G(x))``, i.e. BCE against a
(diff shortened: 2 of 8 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #90, opened 2026-08-09
feat: add SRGANLightning, generator initialisation, and the SRGAN template

YeapJieShen:
⛔ **Do not merge yet — 5th of a 6-PR stack.** Based on `feat/srgan-components`; merge the four below it first.

The SRGAN training loop: `SRGANLightning`, generator initialisation from pretrained weights, adversarial provenance, and the shipped template. Verified by a real GPU run.

Reference: Ledig et al., [arXiv:1609.04802](https://arxiv.org/pdf/1609.04802).

## The loop

`SRGANLightning(SRLightning)` with `automatic_optimization = False`. It **inherits `_forward_lr`/`_forward_sr`/`_step`/`validation_step`/metrics/checkpoint metadata unchanged** and overrides only the optimization surface — so training, benchmarking and inference still route through one pipeline.

Discriminator first, then generator (Goodfellow Algorithm 1), so the generator is updated against the critic that just moved. Goodfellow's `k` (`d_steps_per_g_step`, default 1 = Ledig's 1:1) is spent **across batches** — D every batch, G every k-th — so each discriminator step sees a fresh minibatch, as Algorithm 1 specifies. One generator forward per batch, reused: D's backward runs on `sr.detach()`, and G's flows through a second, fresh D forward. No `retain_graph` anywhere.

`toggle_optimizer`/`untoggle_optimizer` are load-bearing, not decoration. The generator's backward must pass *through* the discriminator to reach the generator, and without toggling it accumulates into D's `.grad` on the way — measured: D carries gradient `[1.241, 1.0]` untoggled versus `[None, None]` toggled, with the generator receiving its gradient either way.

## Units, measured

`global_step` counts optimizer steps, and this module takes **two per batch**, so it advances `N + N//k` per N batches. That is unchanged Lightning behaviour, **not** a quirk of manual optimization — measured over 20 batches: automatic + 1 optimizer = 20, **manual + 1 optimizer = 20**, manual + 2 optimizers = 40. Every other architecture here is unaffected.

The template's comment states each knob's real unit, because they differ: `max_steps`, checkpoint filenames and `every_n_train_steps` are in global steps; `val_check_interval`, `log_every_n_steps` and the hand-stepped scheduler milestones are in **batches**. TensorBoard's default x-axis is batches too, so `sr-weights-10000.pt` corresponds to x=5000 — a 2× offset between a curve and the checkpoint pulled off it.

The template ships the paper's **full** two-phase schedule: `max_steps: 400000` = 2×10⁵ generator iterations, with both networks' LR decaying 1e-4 → 1e-5 at batch 100000.

## Refusals

Five, all loud, because each otherwise mistrains or misloads silently: `cuda_graph` (one capture cannot wrap two alternating optimizers), DDP, `layer_lrs`, a discriminator whose `in_channels` disagrees with the processor, and a discriminator whose declared `hr_input_size` disagrees with the real training crop — validated against the *post-crop* size the discriminator actually receives.

`init_from` refuses **per field** — generator `class_path`, generator `init_args`, `processor.class_path`, `io.output_range`, `io.scale` — each naming the offending field, because weights trained under a different processor or scale produce a model that trains and scores without ever erroring. It also names a `.ckpt`, a discriminator's component `.pt`, and the `--…=null`-becomes-`'None'` CLI trap specifically. It loads only when the run is **fitting**, in `setup` — deliberately not `on_fit_start`, because Lightning restores a `--ckpt_path` resume *between* those two hooks, so loading later would overwrite a resumed generator while the discriminator and optimizer state carried on.

## Testing

**898 passed / 0 skipped / 0 failed** with the GPU visible. `ruff check` and `ruff format --check` clean.

The two load-bearing tests were run five times (5/5 green), and three mutations were verified RED with byte-identical reverts:

| Mutation | Caught by |
|---|---|
| `self.parameters()` instead of `self.model.parameters()` | parameter-disjointness |
| `k` guard removed | k=2 alternation |
| `g_loss = content` (adversarial term dropped) | adversarial-term test |

**That third one is why this section exists.** An earlier revision of these tests passed with the adversarial term dropped entirely — the module would have trained a plain SRResNet while still logging `loss/train/adv`, and scored *better* PSNR for it. The guard now uses a content loss that is identically zero, so the discriminator is the only possible source of a generator gradient, plus a check that halving `adversarial_weight` halves that gradient.

## Smoke run — mechanical, not a quality claim

500 batches from the golden 1M-step SRResNet initialisation, on GPU: 39 min 48 s, ~109 ms/batch steady state.

- CUDA-graph refusal fired at construction, verbatim — not a silent skip.
- `loss/train/d` min **0.216**, max 1.969, zero exact zeros — the critic did not win outright.
- Generator moved: **159/159** learnable tensors, ‖Δ‖₂ = 3.44. `num_batches_tracked` went **1000001 → 1000501**, proving both that the golden init loaded and that exactly 500 batches ran.
- All four loss tags present **and moving** (50 distinct values from 50 points each).
- Rolling windows held independently for all three checkpoint callbacks.
- Both `sr-weights-*.pt` (`kind='sr_model'`) and `d-weights-*.pt` (`kind='component'`) written, neither callback's deletion touching the other's files — the first real exercise of two weight checkpoints with different `attribute` values.

**No image-quality claim is made.** Perceptual gains need ~10⁵ iterations; PSNR is *expected* to fall under an adversarial objective. Any metric from this run is labelled not-evidence in the notes.

## Known issue, documented not fixed

The template's shipped dataloader worker counts **OOM host RAM at the first full validation** — a val worker's float64 `imresize` on a 2040 px DIV2K image, with a discriminator and VGG19 also resident. It fires at the first `val_check_interval`, i.e. hours in, not at step 0. The smoke run completed at 4 train / 0 val workers. The template documents this; no value was changed, because the safe counts on a given box are unmeasured and shipping a guess as though measured would be worse than saying so.

## Stack

```
main → CLI subclass mode → perceptual metrics → rolling checkpoints
     → SRGAN components → THIS PR · SRGAN module + template → docs
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,7 +105,10 @@ Conventional Commit types, with subjects that describe the effect:
 ## Adding a new architecture
 
 You don't write a Lightning module. `SRLightning` already composes an `SRModel` with an
-`SRProcessor`, fed by `SRDataModule`.
+`SRProcessor`, fed by `SRDataModule`. The one exception is a second optimizer: adversarial
+training drives two networks alternately, which Lightning's automatic loop cannot express,
+so `SRGANLightning` subclasses `SRLightning` and owns its own training step. Anything that
+trains one network needs no new module.
 
 1. Subclass `sisr.models.base.SRModel`: `forward`, `self._hparams`, and optionally
    `reset_parameters(**kwargs)` for a paper-faithful init.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,8 +41,10 @@ dependencies = [
     "torch>=2.6",
     "torchvision>=0.21",
     "lightning>=2.4",
-    # 1.9 is the version DISTS was verified present in on this project
-    # (torchmetrics.functional.image.deep_image_structure_and_texture_similarity).
+    # 1.9 is the version DISTS was verified present in on this project. sisr/perceptual.py
+    # caches the backbones directly (torchmetrics.functional.image.dists.DISTSNetwork and
+    # torchmetrics.image.lpip.LearnedPerceptualImagePatchSimilarity.net), so the pin covers
+    # those, not just the functional entry points.
     # Lowering it means testing the lower version, not guessing.
     "torchmetrics>=1.9",
     "jsonargparse[signatures]>=4.27",
--- a/sisr/perceptual.py
+++ b/sisr/perceptual.py
@@ -11,9 +11,19 @@ better than PSNR does, and that is the whole claim.
 DISTS costs nothing — torchmetrics implements it over torchvision's VGG16,
 both already required. LPIPS needs the ``lpips`` package (torchmetrics gates it
 on ``_LPIPS_AVAILABLE``), hence the ``[perceptual]`` extra.
-"""
 
-from collections.abc import Callable
+**Both backbones are memoised here, module-level, keyed by
+``(metric, lpips_net, device, dtype)``.** torchmetrics' *functional* API builds
+a fresh network per call, which dominated the metric entirely: measured on CPU
+at 1x3x256x256, 605.0 ms/call of which 27.3 ms was LPIPS itself, and
+1716.1 ms/call of which 386.8 ms was DISTS itself. One validation cycle of the
+shipped adversarial template scores ~219 images per metric, so that was ~7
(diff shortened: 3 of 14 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #91, opened 2026-08-09
docs: document SRGAN training and the perceptual metrics

YeapJieShen:
⛔ **Do not merge yet — top of a 6-PR stack.** Doc-only; merge the five below it first.

Documents the SRGAN training path and the perceptual metrics, and **revises** the section describing the checkpoint-reload defect that the bottom of this stack fixes.

## The revision matters more than the addition

`README.md` documented — as live behaviour — that `--ckpt_path` rebuilds `eval_config` as the base class and silently drops every default an architecture subclass overrode. **That is fixed**, so the warning, its escalation, and the patch-a-copy-of-the-checkpoint workaround recipe are **deleted**, not softened. A warning that contradicts the code is worse than none.

**The residual is described precisely, because there is one.** A dotted CLI override of a dataclass-typed config field — `--model.init_args.eval_config.ssim_impl=wang`, **no `--ckpt_path` involved** — still applies the value while reverting the object to the base class, so every *other* subclass default reverts with it. Different cause, still open. Scoped to `eval_config`/`training_config`: subclass-typed fields such as `model` merge correctly, verified by counterexample. Two working forms are documented, and `--print_config` distinguishes them (a surviving subclass prints a `class_path`; a reverted one prints a bare mapping).

## Not a breaking change

The subclass-mode CLI was checked rather than assumed: both pre-change templates were instantiated under the new CLI, and jsonargparse treats a `class_path`-less `model:` mapping as `init_args` for the default model class, so **existing configs keep resolving** — subclassed eval configs and all. The old dotted CLI form still parses too. The CHANGELOG says so instead of carrying a false migration warning.

## What else is documented

SRGAN-VGG54 as the reproduced variant; the MSE-trained-SRResNet initialisation and why the paper scopes it to the GAN; the `global_step` accounting (`N + N//k` from two optimizer steps per batch — unchanged Lightning behaviour, **not** a manual-optimization quirk, and no other architecture affected) with each knob's real unit; that PSNR and SSIM get worse **by design**, which is why rolling last-N checkpoints exist; `pip install '.[perceptual]'` for LPIPS while DISTS needs no extra; and that a LPIPS figure compares only within one `lpips_net`, in the same voice as the existing SSIM-convention warning.

**MOS is not reproduced.** Ledig's headline result is a human study. LPIPS and DISTS correlate with it better than PSNR does, and that is the whole claim — they are not a substitute, and the README says so plainly.

No number from the arc's smoke run appears anywhere: 500 batches is a mechanical check, not evidence of quality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,33 @@ notice.
 
 - **SRResNet** and **SRCNN**, both fully wired — model, dataset pipeline, config
   dataclasses, and an experiment template each.
+- **SRGAN** (`sisr/models/srgan/`, `sisr/training/gan_module.py`) — the SRResNet
+  generator trained adversarially against an `SRDiscriminator` critic. `SRGANLightning`
+  drives both optimizers under manual optimization, `AdversarialLoss` supplies the
+  non-saturating objective over the critic's logits, and the shipped template reproduces
+  SRGAN-VGG54. The generator initialises from an MSE-trained SRResNet's bare weights, the
+  paper's recipe, and refuses any weights file whose architecture, processor, output
+  range or scale disagrees with the run loading it. `trainer.global_step` counts
+  optimizer steps, so it advances twice per batch here and nowhere else — see the README.
+- **Subclass-mode CLI** — the top-level `model:` block may now name its Lightning module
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #94, opened 2026-08-10
fix: keep a config's subclass when a dotted CLI override sets one of its fields

YeapJieShen:
## What changed

A dotted CLI override of one field on a config dataclass now sets that field
and changes nothing else. Previously it silently reverted every *other* field
on that config to the base class.

## Why

`--model.init_args.eval_config.crop_border=0` set `crop_border` and, with
nothing logged, also reverted:

- SRResNet's `ssim_impl` from `daala` back to `wang` -- renumbering SSIM under
  a different convention, so the run's numbers stop being comparable to the
  ones it is supposed to be compared against
- SRGAN's `perceptual_metrics` to empty -- removing the only metric family an
  adversarial run can be judged by, since PSNR and SSIM get worse by design

Whether it happened at all depended on how the module's `__init__` annotated
the argument: `SRGANLightning` types `training_config` as its own subclass, so
that one survived, while `eval_config` is typed as the base on `SRLightning`,
so that one did not. That is not a distinction any user could predict from the
config they wrote.

Root cause: jsonargparse registers `is_pure_dataclass` as a subclasses-disabled
selector by default, so these config dataclasses were treated as *closed*
types. A dotted override arrives as a `NestedArg`, which then rebuilds the
whole field from the bare annotation and discards the subclass identity that
had been threaded through correctly.

The CLI now re-enables subclasses for both config families before the parsers
are built. Naming the two base classes covers every architecture's subclass;
dropping the selector by name would have changed parsing for every
dataclass-typed field in Lightning and jsonargparse too.

The dependency floor moves to the first release carrying `subclasses_enabled`,
and to the version this was verified on.

## Evidence

**Bug fix** -- the tests that catch it:

- `tests/test_cli.py::test_dotted_override_keeps_the_config_subclass`
- `tests/test_cli.py::test_dotted_override_keeps_a_base_annotated_training_config_subclass`

Both confirmed **RED** against the unfixed code and **GREEN** after, by
reverting only `sisr/cli.py` and re-running:

| | unfixed | fixed |
| --- | --- | --- |
| `type(eval_config)` | `SREvalConfig` | `SRGANEvalConfig` |
| `ssim_impl` | `wang` | `daala` |
| `perceptual_metrics` | `[]` | `['lpips', 'dists']` |
| `crop_border` (the override) | 0 | 0 |

The first test replaces the characterisation test that pinned the old
behaviour -- replaced, not deleted, so the thing that flips is visible in the
diff.

The second test exists because the first cannot observe half the fix: SRGAN
annotates `training_config` with its own subclass, so no SRGAN config can tell
whether `SRTrainingConfig` is covered. It uses the SRResNet template, where
`SRLightning` annotates the base and `scale` / `init_strategy` exist only on
the subclass -- on a rebuilt base the attribute access itself would raise.

Full suite: **918 passed, 0 skipped, 0 failed**. `--print_config` output is
byte-identical for both shipped templates, so no emitted config changes shape.

## Checklist

## Merge strategy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,11 @@ dependencies = [
     # those, not just the functional entry points.
     # Lowering it means testing the lower version, not guessing.
     "torchmetrics>=1.9",
-    "jsonargparse[signatures]>=4.27",
+    # sisr/cli.py calls set_parsing_settings(subclasses_enabled=...) so a dotted CLI
+    # override of a dataclass config field keeps the architecture's subclass. That
+    # keyword does not exist below 4.50, and 4.50 is the version it was verified on
+    # here. Lowering it means testing the lower version, not guessing.
+    "jsonargparse[signatures]>=4.50",
     "lmdb>=1.4",
     "pillow>=10",
     "numpy>=1.26",
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -35,11 +35,12 @@ from pathlib import Path
 from typing import Any, Literal
 
 import torch
+from jsonargparse import set_parsing_settings
 from lightning.pytorch import Trainer
 from lightning.pytorch.cli import ArgsType, LightningArgumentParser, LightningCLI
 
 from .export import to_onnx
-from .training import SRDataModule, SRLightning
+from .training import SRDataModule, SREvalConfig, SRLightning, SRTrainingConfig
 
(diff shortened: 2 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #95, opened 2026-08-10
fix: bound the per-image gather in imresize so validation stops exhausting host RAM

YeapJieShen:
## What changed

`imresize` now builds its gathered-taps array in slices of the free axis
instead of whole, bounding one image's transient float64 allocation at 64 MiB.

## Why

`_resize_axis` materialised the entire gather at once:
`(out_length, num_taps, free_length, C)` in float64. For a 1536x2040 DIV2K
image at scale 4 that is **286.88 MiB**, per DataLoader worker, per image.

Training never saw it -- a 96x96 crop gathers 0.84 MiB, 341x smaller -- so it
first appeared at a run's first *full-image* validation and killed the run
there, hours in rather than at step 0. The failure was
`numpy._ArrayMemoryError: Unable to allocate 287. MiB for an array with shape
(384, 16, 2040, 3)`, which is exactly this array.

The contraction runs over the tap axis alone, so where the free axis is cut
cannot change any output element's summation. A single slice short-circuits to
the previous code path, which is what keeps the cap off the training path.

## Evidence

**Touches `sisr/imresize.py`** -- byte-equality re-proven:

- `resize()` output **byte-identical** to the previous implementation on 12
  real DIV2K validation images (`np.array_equal` on the final uint8, not a
  tolerance)
- Identical across cap sizes 1 / 4 / 16 / 64 / 256 / 1024 MiB, so the slicing
  boundary is inert rather than incidentally harmless at one setting
- The MATLAB reference byte-equality tests pass (data present on this box, so
  they ran rather than skipped)

**Bug fix** -- the test that catches it:

- `tests/test_imresize.py::test_resize_axis_gather_chunking_is_byte_identical_and_really_splits`
  (both axes)
- It asserts the split output equals the unsplit output **and** that the split
  actually happened -- observed gather sizes are more than one and each within
  the cap. Without the second half the first would pass vacuously under a cap
  that never splits, which is exactly the configuration the training path runs
  in.

**Performance** -- before/after on real data:

| | before | after |
| --- | --- | --- |
| peak gather (exact array size) | 286.9 MiB | **64.0 MiB** |
| full-image validation resize | 157.6 ms | 170.6 ms (+8.2%) |
| 96x96 training crop | 680.1 us | 671.9 us (no measurable change) |

- Medians of 5 isolated runs per variant; warm-up excluded. Within-variant
  spread under 0.7%.
- Measured on: this branch vs `main`'s `imresize.py` extracted verbatim by git,
  real DIV2K validation images, **AC Best-performance overlay**, one variant per
  fresh process, each pinned to a single core.
- The pinning and process isolation are not ceremony: timing both variants in
  one process made whichever ran *second* slower by -15.8% one ordering and
  +37.1% the other, and this CPU is hybrid, so a fresh process landing on a P-
  vs E-core splits a sub-millisecond timing almost exactly 2x. Three earlier
  attempts at this table were wrong before that protocol was adopted.
- The +8.2% is confined to full-image validation (~1.3 s per 100-image pass).
  The training path short-circuits and is unchanged.

**Scope, stated honestly:** this removes the allocation that raised the error,
and a run using the shipped worker counts now completes where it previously
died. But it is ~2.5% of that run's total host-memory demand. The dominant
term is the number of spawned workers, so this bounds a spike rather than
making a marginal configuration safe.

## Checklist

## Merge strategy

### DIFF
--- a/sisr/imresize.py
+++ b/sisr/imresize.py
@@ -67,6 +67,12 @@ import numpy as np
 
 _KERNEL_WIDTH = 4.0  # bicubic support width; MATLAB's a=-0.5 coefficient (OpenCV uses a=-0.75)
 
+# Cap on one gather's float64 footprint in _resize_axis -- see its docstring. Sized
+# well above any training crop (a 96x96 crop at scale 4 gathers 0.84 MiB, so the
+# training path never splits) and well below a full-image validation gather (286.88
+# MiB for a 1536x2040 DIV2K image at the same scale).
+_GATHER_CHUNK_BYTES = 64 << 20
+
 
 def _cubic(x: np.ndarray) -> np.ndarray:
     """MATLAB's bicubic convolution kernel (Keys' family, a=-0.5)."""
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #96, opened 2026-08-12
docs: fix stale claim that the dotted-override config bug is unfixed

YeapJieShen:
P2.11 (a dotted CLI override reverting a subclass-typed config to its base class) was fixed in #94, but README.md's "Config overrides and subclass defaults" section still described it as live and unfixed, including a worked example demonstrating the broken behavior.

Verified against the actual fix: `sisr/cli.py`'s `set_parsing_settings(subclasses_enabled=[SREvalConfig, SRTrainingConfig])`, and the two passing tests `test_dotted_override_keeps_the_config_subclass` / `test_dotted_override_keeps_a_base_annotated_training_config_subclass`.

Doc-only, no code changes.


YeapJieShen (commented):
Human reviewed, all good.

PR #97, opened 2026-08-12
docs: state CI-gate, PSNR/colorspace citations, benchmark provenance, and weights policy

YeapJieShen:
## What

Makes several already-true facts about this project's reproducibility rigor
externally visible in the README — no code changes.

1. **CI-gate visibility.** A sentence next to the `test` badge stating it
   represents a required status check, so a PR cannot merge without the full
   suite passing.
2. **Two peer-reviewed citations in Comparability**, backing the
   convention-inconsistency problems this project already handles explicitly
   (named colorspace functions, the `ssim_impl` split):
   - Keleş, Yılmaz, Tekalp, Korkmaz, Doğan, "On the Computation of PSNR for a
     Set of Images or Video," PCS 2021 (arXiv:2104.14868) — finds no agreed
     convention for aggregating PSNR across a set of images (arithmetic mean
     of per-image PSNR vs. a single PSNR from pooled MSE), diverging up to
     ~2.5 dB.
   - Wang, Chen, Hoi, "Deep Learning for Image Super-resolution: A Survey,"
     IEEE TPAMI 2020 (arXiv:1902.06068), §II-D — finds no accepted best
     practice for which color space/channels to evaluate SR on, with results
     differing by up to 44 dB.

   Also states explicitly which PSNR-aggregation convention this project
   uses (arithmetic mean of independently-computed per-image PSNR values,
   not MSE-pooling), pointing at the exact lines in
   `sisr/training/callbacks.py` and `sisr/training/lightning_module.py` that
   compute and independently confirm it.
3. **Exact benchmark-artifact provenance** — states which Set5/Set14/BSD100
   distribution this project's own numbers are computed against (the EDSR
   authors' `benchmark.tar`), quoting the SHA-256 already recorded in
   `tests/test_imresize.py` so it can't drift out of sync, and notes this
   project's Set14 is the full 14-image variant.
4. **A short `pyiqa` interop note** — its PSNR/SSIM defaults (full RGB;
   Y-channel but full-range YIQ) don't match this project's studio-range
   BT.601 Y-channel convention, so `color_space='ycbcr'` must be passed
   explicitly to reproduce this project's figures.
5. **A "Model weights" policy note** — no checkpoint is released yet; when
   one is, it will be MIT-licensed by default like the code, with an honest
   caveat that whether DIV2K's "academic research purpose only" restriction
   legally propagates to derivative model weights is an unsettled question,
   stated as an open question rather than resolved in either direction.

## Why

This project's flagship differentiator is verified reproducibility —
byte-exact degradation pipelines and named/tested PSNR/SSIM conventions —
which landscape research found apparently unique among comparable SR
toolkits. These facts were already true of the codebase but weren't visible
to a reader of the README. This PR makes them so, and backs the convention
claims with peer-reviewed citations rather than just this project's own say-so.

## Verification

Doc-only change. Read the full file back after editing to confirm no broken
markdown, no duplicate headers, and no contradictions with existing content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@
 [![Python 3.12+](https://img.shields.io/badge/python-3.12%20%7C%203.13-blue)](https://www.python.org/downloads/)
 [![License: MIT](https://img.shields.io/badge/license-MIT-yellow)](LICENSE)
 
+That `test` badge isn't decorative: `test` is a required status check, so a pull
+request cannot merge without the full suite passing.
+
 Reproduce single-image super-resolution papers without rewriting a training loop each
 time. Every architecture runs through one `LightningCLI` entrypoint, and an experiment is
 one self-contained YAML file.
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #98, opened 2026-08-12
ci: skip build/test matrices on doc-only PRs

YeapJieShen:
## What

Skips the expensive build/test matrices on doc-only PRs, without letting a
required status check ever go unreported.

- Adds a `changes` job to both `build.yml` and `test.yml` using
  `dorny/paths-filter` (pinned by commit SHA). It flags `code == 'true'`
  whenever a changed file falls outside a docs allowlist: `**/*.md`,
  `docs/**`, `LICENSE`. Everything else — including `.github/workflows/**`,
  `pyproject.toml`, and `templates/*.yaml` — counts as code.
- Gates the expensive jobs behind `if: needs.changes.outputs.code == 'true'`:
  `build-matrix`, `editable-install` (in `build.yml`) and `test-matrix`,
  `onnx-export` (in `test.yml`). `lint` stays ungated (cheap, single job, and
  not part of the `test` summary job's dependency chain).
- The workflow-level `on:` triggers are untouched — both workflows still fire
  on every PR, so the required `test`/`build` contexts always report.
- Fixes a bug this change would otherwise expose: the `build`/`test` summary
  jobs already run unconditionally (`if: always()`), but their pass/fail
  assertion only accepted a `success` result per dependency. A skipped matrix
  (now possible under the new gate) would have flipped the required check red
  on a doc-only PR. Both assertions now accept `success` **or** `skipped`.

## Why an allowlist, not a denylist

`*.md`, `docs/**`, and `LICENSE` are the only paths confirmed doc-only.
Everything else is treated as code, including paths that look docs-adjacent
but aren't: `templates/*.yaml` is exercised by `tests/test_cli.py`, and
`pyproject.toml` carries both the dependency spec and pytest config.

## Verification

This PR itself touches `.github/workflows/**`, which is deliberately never
doc-only-classified — so its own CI run legitimately runs everything (see
below) and can't by itself demonstrate the skip path.

**1. This PR's own run (code path, unaffected):** every job ran and passed,
including all four matrix legs on both workflows plus `lint`,
`editable-install`, and `onnx-export`. `changes` correctly evaluated
`code == 'true'` since the diff touches `.github/workflows/**`.

**2. Doc-only skip path, proven via a temporary stacked PR:** a throwaway
branch (`ci/verify-doc-only-skip`) was created on top of this feature branch
with a single commit that only appends a line to `CHANGELOG.md`, and a
temporary PR (`#99`) was opened with **this feature branch as its base**
(not `main`) and that throwaway branch as its head. Because the PR's base
already carried this change, `dorny/paths-filter`'s GitHub-API-based diff
(which resolves the real PR base from the event context, not a hardcoded
`main` — confirmed in the job log: *"Fetching list of changed files for
PR#99 ... Detected 1 changed files ... CHANGELOG.md"*) correctly saw a
doc-only diff. Observed job statuses on PR #99:

| Job | Result |
|---|---|
| `changes` (both workflows) | success |
| `build-matrix` (all 4 legs) | **skipped** |
| `editable-install` | **skipped** |
| `test-matrix` (all 4 legs) | **skipped** |
| `onnx-export` | **skipped** |
| `lint` | success (ungated, as designed) |
| **`build`** (required check) | **success** |
| **`test`** (required check) | **success** |

Note: the workflows' `pull_request` trigger only matches `branches: [main]`,
and a `pull_request` event evaluates from the merge-commit context of its
*actual* base branch — so a PR whose base isn't `main` needed that base
branch to (temporarily) also match the trigger before GitHub would run the
workflow at all. This was done with a temporary commit on this branch widening
`branches: [main]` to `branches: [main, ci/skip-doc-only-matrices]`, pushed,
used only to let PR #99 fire, then immediately reverted via a hard reset back
to the prior commit before this PR was left for review — this PR's tip
(`901bf1240da036748eb90f5c451371e0a5f6e065`) has never carried that widening.
The widening commit lived only on the throwaway side of the test and never
appeared in the base↔head diff paths-filter evaluated, so it didn't
contaminate the doc-only signal being tested.

PR #99 was then closed and its branch (`ci/verify-doc-only-skip`) deleted —
it was verification-only and was never intended to merge.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,33 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  # Cheap gate: skips the expensive matrices below on a doc-only PR. Deliberately
+  # an allowlist, not a denylist — anything not explicitly listed here (including
+  # this workflow file itself) is treated as code and runs the full matrix.
+  changes:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d  # v4.0.3
+        id: filter
+        with:
+          predicate-quantifier: some-with-excludes
+          filters: |
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!LICENSE'
+
   build-matrix:
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     strategy:
@@ -50,6 +76,8 @@ jobs:
   # CONTRIBUTING.md gives contributors who skip uv exactly that command. One
   # OS/Python leg is enough — this checks build-backend behaviour, not the matrix.
   editable-install:
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #100, opened 2026-08-12
chore: add mypy static type checking and a py.typed marker

YeapJieShen:
## What

Adds `mypy` static type checking, gated in CI, plus a PEP 561 `sisr/py.typed` marker.

## Checked now

`mypy sisr` (see `[tool.mypy]` in `pyproject.toml`) actually type-checks:

- `sisr/colorspace.py`, `sisr/ssim.py`, `sisr/imresize.py`, `sisr/cache.py`
- `sisr/losses/` (whole package)
- `sisr/datasets/` (whole package)
- `sisr/processors/` (whole package)
- `sisr/export.py`, `sisr/perceptual.py`
- the model network-definition files (`sisr/models/base.py`, `sisr/models/*/model.py`, `sisr/models/srgan/discriminator.py`) and, since they turned out to be plain, statically-shaped dataclasses rather than jsonargparse-dynamic wiring, the model `config.py` files too (`sisr/models/srcnn/config.py`, `sisr/models/srresnet/config.py`, `sisr/models/srgan/config.py`)

## Ignored, and why

`[[tool.mypy.overrides]]` blanket-ignores (`ignore_errors = true`):

- `sisr.training.*` (the whole subpackage) — Lightning's hook system
- `sisr.cli` — jsonargparse's dynamic CLI/subclass resolution

Both make static typing resist useful checking without a much larger, separate effort — per the agreed scope for this pass.

Also `ignore_missing_imports = true` for `torchvision.*` and `tqdm.*`, neither of which ships a `py.typed` marker or inline types.

## Real fixes made (no suppressions needed)

mypy raised 20 real errors across 9 files in the checked set; all were fixed for real — no `# type: ignore` was needed anywhere:

- Missing return/parameter type annotations (`sisr/cache.py`, `sisr/models/base.py`, `sisr/models/srcnn/model.py`, `sisr/models/srresnet/model.py`, `sisr/datasets/base.py`)
- Two narrow `assert`s in `sisr/cache.py` making explicit an invariant the code already relied on at runtime (`_kernel32` is only `None` off-Windows; `_heartbeat_thread` is always set alongside `_heartbeat_stop`)
- `sisr/losses/vgg.py`: class-level type annotations for attributes set via `object.__setattr__`/`register_buffer`, which otherwise fall back to `nn.Module.__getattr__`'s `Tensor | Module` union
- `sisr/losses/composite.py`: an `assert` making explicit that the accumulator loop always runs at least once (enforced by the constructor's non-empty-terms check)
- `sisr/datasets/base.py` + `sisr/datasets/predict.py`: widened `SRDataset.__getitem__`'s abstract return type to `tuple[Tensor, Tensor] | Tensor` — `PredictDataset` genuinely has no HR pair to return (real inference images, no ground truth), so the narrower base signature was actually wrong, not just under-annotated
- `sisr/models/srcnn/model.py`: `reset_parameters` now takes `**kwargs: Any` (matching the base class's documented polymorphic-call contract) and guards `Conv2d.bias` being `None`
- `sisr/perceptual.py`: added a `@runtime_checkable` structural `Protocol` (`_ResettableMetric`) instead of asserting against `torchmetrics.Metric` directly — the latter would have broken the test suite's duck-typed stand-in for the LPIPS backbone

Also pinned `platform = "win32"` in `[tool.mypy]`: `sisr/cache.py`'s Windows liveness check uses `ctypes.WinDLL`, which typeshed only exposes under that platform, so the exact same code type-checks clean on Windows and fails with `attr-defined` on Linux/macOS without the pin. The new `typecheck` CI job accordingly runs on `windows-latest` (already a supported leg via `test-matrix`), not `ubuntu-latest`.

## Packaging

- `sisr/py.typed` added (empty PEP 561 marker).
- Verified hatchling's existing `packages = ["sisr"]` wheel config already includes it — built a real wheel locally and inspected its contents rather than assuming.
- `tests/test_packaging.py::test_wheel_ships_py_typed` now asserts this directly: it builds a wheel via hatchling's PEP 517 `build_wheel()` hook and checks `sisr/py.typed` is actually in it, not just that the config that's supposed to produce it looks right.

## Acceptance evidence

- `mypy sisr` — clean, 0 errors, 44 source files checked.
- `ruff check .` and `ruff format --check .` — clean.
- Full test suite (`pytest`, this data-less worktree): **800 passed, 3 skipped** (skips are the pre-existing data-gated byte-equality tests, not new).
- Wheel built locally via `hatchling.build.build_wheel()` and confirmed to contain `sisr/py.typed`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,6 +65,36 @@ jobs:
       - name: Ruff format check
         run: ruff format --check .
 
+  # Type-checks only the modules [tool.mypy]'s overrides don't blanket-ignore
+  # (Lightning's hook system and jsonargparse's dynamic CLI/subclass
+  # resolution make sisr/training/, sisr/cli.py resist useful static typing
+  # without a large, separate effort). Installs the dev extra rather than a
+  # bare `mypy` so it type-checks against the same third-party stubs/versions
+  # the rest of CI runs against.
+  # windows-latest, not ubuntu-latest: sisr/cache.py's ctypes.WinDLL Windows
+  # liveness check is genuine platform-specific code, guarded by a runtime
+  # sys.platform == "win32" check -- but typeshed's ctypes stubs only expose
+  # WinDLL under that same platform, so mypy raises attr-defined errors for it
+  # when its own --platform is anything else. test-matrix already treats
+  # windows-latest as first-class, so this matches, not adds, a supported OS.
+  typecheck:
+    runs-on: windows-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+      # See the lint job for why this pins an exact patch version.
+      - uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          cache-suffix: typecheck
+      - name: Install
+        run: uv pip install --system -e ".[dev]"
+      - name: Mypy
+        run: mypy sisr
+
   test-matrix:
     needs: changes
     if: needs.changes.outputs.code == 'true'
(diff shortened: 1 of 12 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #103, opened 2026-08-15
fix: put diagnostic callbacks on the same step axis as every other metric

YeapJieShen:
## What

`GradNormLogger` and `WeightHistogramLogger` gated `log_every_n_steps` on `trainer.global_step`, which counts **optimizer** steps. Under manual optimization (the adversarial module steps D and G per batch) that is 2× the batch count, so:

- the configured cadence meant a different number of batches depending on the training paradigm, and
- `GradNormLogger` sampled against one counter while emitting through `self.log`, which plots on the other.

`WeightHistogramLogger` carried a second, sharper instance: it stamped `add_histogram` on the **raw** TensorBoard experiment with `global_step`, so its histograms would land at twice the step of the loss curves they are read against. That is the same defect #101 fixed for benchmark images, at a site #101 never touched.

Both now read `_logger_step`, the helper #101 introduced.

## The unit is batches

Deliberate choice, not a typo fix. Batches mean `log_every_n_steps` says one thing regardless of paradigm and matches the axis the metric is plotted on. The alternative — optimizer steps — has a real argument (a grad-norm trace arguably wants per-optimizer-step resolution under adversarial training) but costs a config value whose meaning changes per model. Accepted cost: under a two-optimizer paradigm the trace samples half as often per optimizer step.

## Evidence

Three tests, each **proven RED against pre-fix `main`** before the fix landed:

- `test_grad_norm_logger_cadence_counts_batches_not_optimizer_steps`
- `test_weight_histogram_logger_cadence_counts_batches_not_optimizer_steps`
- `test_weight_histogram_logger_writes_on_the_batch_counted_axis_not_global_step` — fails as `assert 400 == 200`, the same signature #101's own proof used

Each cadence test asserts **both** directions (batch-axis on / optimizer-axis off, and the reverse); a stub whose two axes agree, or a single direction, passes under a gate reading either axis. The histogram probe uses a real `TensorBoardLogger` rather than a mock, because the callback finds it through an `isinstance` check that would filter a mock out and make every assertion pass vacuously.

Additionally verified by **isolated mutation**: reverting only the `add_histogram` axis while leaving the gate fixed kills exactly the axis test and leaves the two cadence tests green.

Seven pre-existing tests had trainer stubs carrying only `global_step`; they now use the shared two-axis fixture, with values chosen so the axes disagree rather than agree.

## Verification

- Full suite with real data: **923 passed / 2 skipped** (920 before; the 2 skips are the Windows-only cache liveness paths). New tests re-run 5× for stability.
- `ruff check` + `ruff format --check` clean; `mypy sisr` clean.

## Blast radius

None to existing data. `GradNormLogger` is wired only in the SRCNN/SRResNet templates, where automatic optimization makes `global_step` equal the batch count; `WeightHistogramLogger` is wired in no template. The bug was latent on every shipped config.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -513,8 +513,11 @@ class GradNormLogger(Callback):
     after the backward pass and logs it as the ``"diag/grad_norm"`` scalar.
 
     Args:
-        log_every_n_steps (int): Compute and log every *n* training steps.
-            Defaults to ``100``.
+        log_every_n_steps (int): Compute and log every *n* **batches**.
+            Defaults to ``100``. The unit is batches, not optimizer steps, so
+            the configured cadence means the same thing under automatic and
+            manual optimization — and matches the axis the emitted scalar is
+            plotted on. See :func:`_logger_step`.
     """
 
     def __init__(self, log_every_n_steps: int = 100):
@@ -533,7 +536,7 @@ class GradNormLogger(Callback):
             trainer: The trainer instance.
             pl_module: The model being trained.
         """
-        if trainer.global_step % self.log_every_n_steps != 0:
+        if _logger_step(trainer) % self.log_every_n_steps != 0:
             return
 
         grad_norms = [p.grad.detach().norm(2) for p in pl_module.parameters() if p.grad is not None]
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #105, opened 2026-08-16
refactor: give the two SR checkpointers one shared module

YeapJieShen:
From the architecture review — the only candidate rated **LOW risk**, which under the standing unattended-session rule is the only one an agent may implement rather than propose. The other five are filed as proposals.

## Problem

`SRCheckpoint` and `SRWeightsCheckpoint` were siblings under `ModelCheckpoint` rather than parent and child. One concept — "an SR checkpointer: monitored-or-rolling, sanitised filename, validated monitor" — was therefore spread across a free function, a mixin, and two classes:

- The filename construction and the rolling-mode `save_top_k = -1` rule were written **twice**, verbatim.
- `_validate_monitor_metric` took **four** arguments to perform one check, one of which (`label`) existed solely so the error could name its caller — a value both callers already knew about themselves.
- Both docstrings stated the same justification: *"siblings … not parent/child, so this cannot be reused via `super()`"*. That is a consequence of the chosen shape, not a reason for it — `SRWeightsCheckpoint._save_checkpoint` never calls `super()._save_checkpoint`, so inheriting pulls in none of the full-payload write it exists to avoid.

## Change

One `_SRCheckpointBase` owns construction, the rolling rule, the window bookkeeping and monitor validation. Each subclass keeps only its defaults and, where they differ, its payload override. Two adapters over one module.

The validation call goes from four arguments to one, and the caller's name is now derived via `type(self).__name__` rather than passed as a literal, so the message and the class cannot drift.

## Behaviour is unchanged

Same filenames, same rolling rule, same `state_key`, same YAML `class_path`/`init_args` surface. No seam moves; the metric path is untouched.

## Verified by mutation, not by reading

- **Removing the rolling-mode `save_top_k = -1` rule kills 18 tests** — that invariant is well guarded.
- **Deriving the class name killed nothing.** No test pinned the class name in a message users actually read, and this change alters how that name is produced. That gap is closed here by a new test, parametrised over **both** classes — a single-class check still passes when the name is hard-coded to that one class, and I confirmed the mutation fails on the `SRWeightsCheckpoint` arm specifically.

Tests that were parametrised over both classes purely to exercise duplicated code now target the base once.

## Checks

- Full suite: **804 passed / 5 skipped** in the worktree (data-less, so the byte-equality and real-image SSIM cases skip; `--collect-only` is 807 vs the main checkout's 925 — the difference is entirely `test_ssim.py` cases parametrised over image sets that aren't present).
- `ruff check` + `format --check` clean; `mypy sisr` clean, 44 files.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -616,9 +616,9 @@ def _validate_monitor_metric(
 ) -> None:
     """Raise if ``monitor`` names no logged metric, or is monitored in the wrong direction.
 
-    Shared by :class:`SRCheckpoint` and :class:`SRWeightsCheckpoint`'s ``setup`` —
-    they are siblings under :class:`~lightning.pytorch.callbacks.ModelCheckpoint`,
-    not parent/child, so this check cannot be reused via ``super()`` across them.
+    Implementation behind :meth:`_SRCheckpointBase._validate_monitor`, which is
+    how every caller reaches it — ``label``/``monitor``/``mode`` are all ``self``
+    there. Kept as a module-level function only because the rule is pure.
 
     The direction half exists because both classes default to ``mode='max'``,
     which is right for PSNR/SSIM and exactly inverted for LPIPS/DISTS — the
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #106, opened 2026-08-16
fix: score benchmark PSNR with the same reduction as validation

YeapJieShen:
Found by the architecture review; filed as its own entry because it is a concrete latent defect, not a refactor.

## The defect

`SRLightning._mean_psnr` computes per-image PSNR then a batch mean (`dim=(1,2,3)`), and its docstring states that reduction is deliberate — *"the deviation a stateful `PeakSignalNoiseRatio` with default `dim` would introduce"*.

`BenchmarkImageLogger._collect_batch` instead called `torchmetrics` directly with no `dim`, i.e. `dim=None`, pooling the whole batch into one PSNR.

The comment **immediately above that call** already declares the invariant it breaks:

> *"…reaching into it here (rather than calling torchmetrics directly) keeps this path and `validation_step`'s unable to disagree on what `ssim/...` tags under the same name."*

SSIM honoured that. PSNR did not.

## Not currently wrong — which is why it needed finding

`_collect_batch` slices `sr[i:i+1]`, so every call is a batch of one, where both reductions coincide exactly. **Today's published per-set numbers are correct.** The hazard is that:

- nothing states or tests the batch-of-one precondition;
- any future edit to `_mean_psnr`'s reduction is silently not inherited;
- batching that loop is an obvious optimisation — it currently runs one generator forward *per image* — and doing it would quietly change every `psnr/{set}/{key}` number with no test failing.

## Published figures do not move

Scored the converged 1M-step SRResNet checkpoint through the `test` path before and after, same config, same weights:

| metric | pre-fix | post-fix | delta |
|---|---|---|---|
| psnr/Set5/Y | 32.057113647 | 32.057205200 | +9.2e-05 |
| psnr/Set14/Y | 28.489923477 | 28.489831924 | −9.2e-05 |
| psnr/BSD100/Y | 27.514175415 | 27.514192581 | +1.7e-05 |

Largest deviation **9.2e-05 dB** — float summation order (`dim=None` does one MSE-then-log over all elements; `dim=(1,2,3)` does per-image PSNR then a one-element mean). Note SSIM shifted by ~2e-06 across the same pair **despite being untouched by this change**, which places run-to-run cuDNN autotune nondeterminism in the same residual. Four orders of magnitude below the precision any paper comparison is quoted at.

## Tests

- `test_benchmark_psnr_routes_through_the_module_not_torchmetrics_directly` — **proven RED on pre-fix `main`**. It stubs `_mean_psnr` to return a sentinel and asserts the buffered sample carries it, so it constrains the *result*, not the call mechanism; a callback computing its own PSNR cannot produce the sentinel. Re-verified by mutation: reverting the one-line fix kills it.
- `test_the_two_psnr_reductions_genuinely_differ_above_batch_size_one` — pins the dependency behaviour that makes the routing matter. Worth noting its first draft **passed under both reductions**, because both images were given the same error, so their per-image PSNRs were equal and pooling agreed with the mean. The fixture now uses differing errors and the test documents that trap.

Full suite 806 passed / 5 skipped in the worktree; `ruff` and `mypy` clean.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -20,7 +20,6 @@ from weakref import proxy
 import lightning
 import torch
 import torch.nn.functional
-import torchmetrics.functional
 import torchvision
 from lightning.pytorch.callbacks import BasePredictionWriter, Callback, ModelCheckpoint
 from lightning.pytorch.utilities.exceptions import MisconfigurationException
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #108, opened 2026-08-16
refactor: give the scoring path one owner

YeapJieShen:
> **Stacked on #107** — base is `test/pin-metric-tags`, not `main`. Review #107 first; this PR's diff is only the refactor.

## Why

There was no module answering *"score an aligned `(sr_rgb, hr_rgb)` pair under this eval config"*. `SRLightning` exposed four private methods as a de-facto interface and `BenchmarkImageLogger` reached through all of them. The callback's own comment conceded it:

> Still a private reach: the colorspace split has no public seam, and duplicating it here would recreate a divergence this design removed.

Four decisions had no single home — the **crop-border slice** (written twice), the **colorspace split**, the **PSNR reduction**, and which SSIM `ssim_impl` names. The tag grammar `{family}/{scope}/{key}` was rebuilt from f-strings in **four** places, so nothing stopped `psnr/val/Y` and `psnr/Set5/Y` drifting apart.

This is INIT.46 candidate 1, and its first symptom already shipped as its own fix (#106) — the benchmark path had in fact been "tidied" into a different PSNR reduction, and agreed with validation only because it slices one image at a time.

## What

New `sisr/training/scoring.py`:

- `SRScorer` — built from an `SREvalConfig` **alone**. No `SRLightning`, no `Trainer`, no datamodule, so scoring is testable on two tensors and a config.
- `metric_tag(family, scope, key)` — the grammar's only implementation. Callers supply the scope (`"val"`, or a dataset name) and nothing else.
- `Scores` — the values, untagged, with `.tagged(scope)` to flatten.

Clean break, no shims: the four private methods are **gone**, and both call sites share one scorer object. Callers keep what is genuinely theirs — `prog_bar` is a display choice, so it stays in `validation_step`.

## Evidence that behaviour is unchanged

This is the part that matters for a path carrying the flagship claim, and it is verified rather than asserted:

**The golden-value contract from #107 passes with identical values before and after.** It pins every emitted tag and every number to 9 significant figures under **both** `ssim_impl` values — including the PSNR reduction, which its fixture is specifically built to detect. That is the whole reason #107 landed first.

Existing coverage also holds unchanged, including `test_collect_batch_metric_values_match_pre_change_host_copy_first_ordering`, which pins the benchmark path's values from before an earlier change.

## Test changes

Mechanical — the same assertions retargeted from the private methods to the scorer. One is a genuine improvement: the benchmark stub module now carries a **real** `SRScorer` over a real `SREvalConfig` instead of a mocked one, because stubbing the scorer would let the callback stop scoring entirely without failing.

## Verification status

`ruff` and `mypy sisr` clean. Affected suites green (237 passed across `test_callbacks.py`, `test_lightning_module.py`, `test_scoring_contract.py`).

**The full suite has not been run against this branch yet** — a GPU screen is running and the standing convention here is not to run the suite beside one. It will be run once the screen finishes and the result posted as a comment before this is ready to merge.

## Not done here

Candidate 6 (`metadata.py`'s read side) stays deliberately held — one adapter is a hypothetical seam.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -341,10 +341,11 @@ class BenchmarkImageLogger(Callback):
         slices — mirroring the primary-val-loader path in
         ``SRLightning.validation_step`` — so only the scalar ``.item()``
         results ever leave the GPU. SSIM and perceptual scores go through
-        ``pl_module._mean_ssim``/``pl_module._mean_perceptual`` rather than calling
-        ``torchmetrics``/``perceptual_score`` here directly, so this callback
-        cannot silently diverge from ``validation_step`` on which SSIM
-        implementation or perceptual backbone a given ``eval_config`` means. When
+        ``pl_module.scorer`` — the same object ``validation_step`` uses — rather
+        than calling ``torchmetrics``/``perceptual_score`` here directly, so
+        this callback cannot silently diverge from ``validation_step`` on the
+        crop, the reductions, or which SSIM/perceptual backbone a given
+        ``eval_config`` means. When
         *should_log_images* (or
         ``self.log_per_image_metrics``), exactly one host transfer per image
         (``lr_img[i]``/``sr[i]``/``hr_img[i]`` each ``.cpu()`` at most once)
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

YeapJieShen:
Full suite against the top of the stack: **933 passed, 2 skipped** in 3m40s (929 → 933 collected; the 6 new tests, minus 2 parametrised cases counted once). The 2 skips are the Windows-specific cache liveness paths, so the data-gated `imresize` byte-equality tests ran and passed.

`ruff` clean, `ruff format --check` clean, `mypy sisr` clean.

This was deferred earlier because a GPU screen was running and the standing convention here is not to run the suite beside one; the screen has since finished.

PR #110, opened 2026-08-19
refactor: give the setup probe's sampling discipline one owner

YeapJieShen:
## What

`SRLightning.setup()` probes one real `(lr, hr)` sample to catch a model/dataset contract mismatch before training starts. That read carries two guarantees, both load-bearing and neither visible at the call site:

- **RNG transparency** — `TrainDataset.__getitem__` draws its crop with `random.randint`, so an unguarded probe consumes draws from the sequence the training loop uses and shifts every seeded crop after it.
- **Pickle-clone reads** — the LMDB-backed datasets cache an unpicklable `lmdb.Environment` on first access. Reading the live instance poisons it for any later `num_workers > 0` loader, and the crash surfaces inside torch or Lightning, far from the cause.

Both lived inside `setup()`: the RNG snapshot inline in the method body, the pickle clone in a sibling staticmethod, and neither reachable without a `LightningModule`, a `Trainer` and a datamodule.

`_extra_probe` is the evidence that this was the wrong shape. Its docstring says it exists so a subclass can validate against real data *without repeating* the discipline — which is another way of saying the discipline could be **avoided but never reused**.

## How

New `sisr/training/probe.py` owns both guarantees. `probe_pair` is a **context manager**, not a function, and that is the design point: the guard covers the caller's whole use of the sample — the lazy train read and any subclass hook included — and there is no way to obtain a sample from outside the guarded region. What was a convention a reviewer had to remember is now the shape of the API.

`setup()` keeps only the checks. It depends on nothing but `random`, `pickle` and duck-typed reads — no Lightning, no `SRLightning`, no torch beyond a type annotation.

## Behaviour

Unchanged, deliberately including the two subtle parts:

- index 0 is read **at most once** when the train dataset is both the probe source and the train-shape source (`train_lr()` is lazy *and* memoised, which is what preserves this);
- the fall-back-to-live-object path for a dataset something else already opened.

`sisr/training/lightning_module.py`: **-119 / +18**.

## Verification

- Full suite in the worktree: **823 passed, 5 skipped** (all five are the documented data-gated ones — no `data/` in a worktree).
- `ruff check` + `ruff format --check` + `mypy sisr` clean.
- Checkout asserted by exact `sisr.__file__` equality before every run.

**Mutation-verified against the new test file alone**, so the coverage is real and not inherited from the existing end-to-end tests:

| mutation | result |
|---|---|
| drop the RNG restore | **3 tests fail** |
| read the live object instead of a clone | **1 fails** |
| never reuse the probe sample for `train_lr` | **1 fails** |
| inert control edit | **0 fail** (correct — the harness isn't just failing on any change) |

The new tests state each invariant against a dataset built to exercise it. The existing end-to-end tests are valuable and stay, but they can only assert RNG transparency *through* SRResNet's crop — move that crop to a private generator and they would pass while asserting nothing.

### DIFF
--- a/sisr/training/lightning_module.py
+++ b/sisr/training/lightning_module.py
@@ -9,8 +9,6 @@ itself. Optimizer / LR scheduler are wired in from top-level YAML keys by
 """
 
 import dataclasses
-import pickle
-import random
 from typing import Any
 
 import lightning
@@ -25,6 +23,7 @@ from ..processors import SRProcessor
 from .config import SREvalConfig, SRTrainingConfig
 from .cuda_graph import CUDAGraphStep
 from .metadata import build_metadata
+from .probe import probe_pair
 from .scoring import SRScorer, metric_tag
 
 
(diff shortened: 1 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #111, opened 2026-08-19
fix: stamp checkpoints with the step axis the metrics are logged on

YeapJieShen:
## The defect

A checkpoint records which step produced it so a reader can line it up against a curve. It was recording a number that **appears on no curve**.

Lightning fills the `{step}` filename placeholder — and our metadata — from `trainer.global_step`, the *optimizer* count. Every `self.log` metric lands on `_batches_that_stepped`, which Lightning keeps batch-counted deliberately for loggers. Those differ **two ways at once**:

- **by one, always** — `global_step` counts *completed* steps while the logger stamps the index of the step just completed;
- **by the optimizers-per-batch factor** under manual optimization — an adversarial module steps D and G per batch, so 2×.

Measured on the real runs:

| run | max metric step | max checkpoint name | ratio |
|---|---|---|---|
| `SRGAN/golden` | 199,999 | `sr-400000` | **2.00×** |
| `SRGAN/golden-mse` | 199,999 | `sr-400000` | **2.00×** |
| `SRResNet/golden` | 999,999 | `sr-1000000` | off by one |
| `SRCNN/baseline` | 1,131,999 | `sr-1000000` | off by one |

Weights and metrics were both correct throughout. Only the stamp was wrong.

## The fix

`_SRCheckpointBase._monitor_candidates` substitutes the batch axis. Lightning routes filename formatting, top-k bookkeeping and the `last`/`every_n` paths through that one method, so a single override keeps them consistent with each other.

The metadata records **both** counters. `global_step` keeps meaning the optimizer count — that is what it factually is, and it is the name Lightning's own checkpoint payload uses for the same quantity, so redefining it would put our field in direct contradiction with the one beside it. The batch axis is added as `batch_step`, which also means an older reader that ignores the new key still reads a true value. Carried through the `.pt` path, the `.ckpt` path (read from the checkpoint's own persisted loop state, the same discipline `global_step`/`epoch` already follow) and ONNX export.

## ⚠️ This is not invisible on automatic-optimization runs

Names shift by **one** — because that off-by-one was itself part of the defect. A 1e6-batch run now ends at `sr-999999`, which is a real point on the curve, where `sr-1000000` was not. **Existing checkpoints are untouched**; only new saves are affected.

## Verification

Proven **end-to-end against the actual property**, not against a counter: a real fit with a real logger, intersecting the set of stamped steps with the set of logged steps.

| | stamped steps | logged steps | overlap |
|---|---|---|---|
| with the fix | 3, 7, 11, 15, 19 | 3, 7, 11, 15, 19 | **5/5** |
| override disabled | 4, 8, 12, 16, 20 | 3, 7, 11, 15, 19 | **0/5** |

Zero overlap without the fix — not a near-miss.

- Full suite: **833 passed, 5 skipped** (all five documented data-gated; no `data/` in a worktree).
- `ruff check` + `ruff format` + `mypy sisr` clean.
- Checkout asserted by exact `sisr.__file__` equality.
- RED proven before the fix on all five new behavioural tests.

One test deliberately guards against a *wrong* fix: `test_checkpoint_name_introduces_no_scaling_of_its_own` fails if anyone "fixes" the 2× by dividing `global_step` by an assumed optimizers-per-batch factor instead of reading the batch counter.

### DIFF
--- a/sisr/export.py
+++ b/sisr/export.py
@@ -99,6 +99,7 @@ def to_onnx(
     _require_onnx()
 
     global_step: int | None = None
+    batch_step: int | None = None
     epoch: int | None = None
     monitor: str | None = None
     monitor_value: float | None = None
@@ -111,6 +112,10 @@ def to_onnx(
         global_step = checkpoint.get("global_step")
         epoch = checkpoint.get("epoch")
         ckpt_training = checkpoint.get("sisr_meta", {}).get("training", {})
+        # The batch axis has no top-level key of its own -- Lightning writes only
+        # `global_step` there -- so it comes from the sisr envelope, and is absent
+        # (None) on any checkpoint written before that field existed.
+        batch_step = ckpt_training.get("batch_step")
         monitor = ckpt_training.get("monitor")
         monitor_value = ckpt_training.get("monitor_value")
 
@@ -164,6 +169,7 @@ def to_onnx(
         module,
         file_path,
         global_step=global_step,
+        batch_step=batch_step,
         epoch=epoch,
         monitor=monitor,
         monitor_value=monitor_value,
@@ -175,6 +181,7 @@ def _write_metadata_props(
     file_path: str | PathLike,
     *,
     global_step: int | None,
+    batch_step: int | None,
     epoch: int | None,
     monitor: str | None,
     monitor_value: float | None,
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

Issue #121, opened 2026-08-26
Re-measure perf floors after the structural refactor


YeapJieShen:
> **Blocked by #113 and #114.** Both dismantle the structure these numbers would describe. A perf
> figure measured against code about to be deleted outlives the code it describes — that is the
> failure this whole re-baseline exists to correct.

## Context

The entire performance corpus was measured against the current code structure and the previous
platform. After the graph path is retired (#113) and the modules are reorganised (#114), most of it
describes code that no longer exists.

The reference config ships **plain eager fp32** as of #113 — bit-identical by definition, no
runtime toolchain dependency. This issue decides whether anything should displace that.

## Scope

Re-baseline **only what a live decision depends on** — not a full re-run of every historical cell.

### 1. Backend sweep, scope-matched

The previous comparison was unfair: graph capture covered the whole training step while
`compile_backend` wrapped the model only. Now that the graph path is gone, establish honest
numbers across **every backend available**, not just inductor:

- `inductor` (the presumed winner)
- `cudagraphs`
- `aot_eager`
- `eager` as the control
- any other backend `torch._dynamo.list_backends()` reports on this box

Report warmup-excluded medians with full provenance. Include compile/warmup cost separately —
a backend that wins per-step but costs minutes to warm may still lose on a short run.

### 2. Bit-equality, per backend

**This is the decision-critical measurement.** Graph capture was verified bit-identical to eager.
Compilation fuses operators and may reassociate floating-point ops.

For each backend, compare output tensors against eager fp32 and report the exact deviation. A
backend is only eligible to become the reference default if it is bit-identical, or if its
deviation is argued and accepted explicitly.

### 3. SRCNN cells

Same matrix shape, SRCNN config — it is kernel-launch-bound where SRResNet is not, so the ranking
may differ.

## Acceptance

- [ ] Scope-matched ms/step for every available backend, medians, provenance recorded
- [ ] Bit-equality verdict per backend against eager fp32, with the tolerance stated
- [ ] Compile/warmup cost reported separately from steady-state
- [ ] An explicit recommendation: keep eager as the reference default, or displace it — with the
      numerics argument, not just the speed one
- [ ] Superseded figures marked as superseded, never left standing

YeapJieShen:
## Substantially answered

`Snapshot @ 8763979`. Real-run rates re-measured now that the structure this issue was waiting on has
landed. Stripped reference config, callbacks off, three 240 s windows each, cadence 20 so every
[...]

PR #152, opened 2026-08-26
refactor: retire the hand-maintained CUDA graph capture path

### LINKED ISSUE #113: Retire the hand-maintained CUDA graph capture path
## Context

`SRLightning` can capture `{zero_grad, forward, loss, backward}` into a CUDA graph and replay it
(`training_config.cuda_graph`, default **off**). It exists because an SRCNN training step is
kernel-launch-bound: 78 `cudaLaunchKernel` calls per step, CPU launch time equal to step wall
time, GPU under 30% utilisation.

## Proposal

Delete the hand-maintained capture path and get the same speedup from `torch.compile` backends.

This is **not** "give up CUDA graphs" — inductor's `reduce-overhead` mode builds its own cudagraph
trees. It is "stop maintaining our own."

## Why

**Coverage shrinks while cost stays fixed.** Capture is refused outright for the adversarial
paradigm (two optimizers, manual optimization), and refused for mixed precision, distributed
training, gradient accumulation, and non-CUDA devices. Every architecture family added from here
inherits the refusal rather than the benefit.

**Maintenance surface.** ~330 lines in the graph module, ~180 more in the Lightning module spread
across four hooks and two state flags, and 82 test references.

**Two of the project's standing traps exist only because of this feature**, both in the
"silently corrupts training forever" class — a `zero_grad(set_to_none=True)` under a live graph
freezes training while the loss curve keeps moving, and per-term loss buffers must be written in
place or one interleaved eager forward strands every per-term tag. Retiring the feature retires
both traps.

## What it costs

`Snapshot @ 5abcbdf` — SRResNet, batch 16, 96×96, ×4, 350 steady batches per cell, Linux:

| configuration | ms/step |
| --- | --- |
| graph capture, fp32 | 25.7 |
| eager fp32 | 35.5 |
| inductor `reduce-overhead`, fp32 | 26.3 |
| inductor `reduce-overhead`, bf16 | 16.0 |

The two fp32 rows are **not scope-matched**: capture covers the whole training step, while
`compile_backend` wraps the model only — loss, backward and optimizer stay eager. So "inductor
fp32 loses" means *compiling the model alone does not beat capturing the whole step*, not
*inductor cannot beat capture*. Establishing the scope-matched number is part of the follow-up
perf pass, not of this change.

## Open question, deliberately not a blocker

Graph capture is verified bit-identical to eager. Inductor fuses operators and may reassociate
floating-point ops; whether it stays bit-identical here has never been measured. That matters
only for the reference reproduction config, and it is sequenced **after** this change as part of
the post-refactor perf pass — measuring it now would produce a number describing a structure
about to be dismantled.

## Acceptance

- [ ] Graph module and its Lightning-side contract deleted
- [ ] Reference config migrated to the compile path or to plain eager
- [ ] The two graph-lifetime traps removed from the project notes
- [ ] Suite green, no behaviour change for runs that never opted in

YeapJieShen:
Closes #113.

## What changed

The hand-maintained CUDA graph capture path is gone: the capture module, the
`training_config.cuda_graph` field, and the four Lightning hooks plus two state
flags that existed to keep replay sound. `compile_backend` is untouched and still
opt-in. A run that never set `cuda_graph` behaves identically.

## Why

Coverage was shrinking while cost stayed fixed. Capture was refused outright for
the adversarial paradigm, and for mixed precision, distributed training and
gradient accumulation — so each new architecture family inherited the refusal
rather than the benefit. `torch.compile`'s backends cover the same ground without
this project owning the correctness argument; inductor's `reduce-overhead` builds
its own cudagraph trees.

It also retires the two "silently corrupts training forever" hazards the feature
created, both able to freeze training while `loss/train` kept moving.

The reference SRResNet config now ships **plain eager fp32** — bit-identical by
definition, with no runtime toolchain dependency. Honest scope-matched per-backend
numbers are #121's job, deliberately sequenced after this: the previous comparison
was unfair (capture covered the whole step, `compile_backend` wraps the model only),
and measuring against a structure about to be dismantled is the failure this
re-baseline exists to correct.

## Evidence

**Suite** — `911 passed, 2 skipped` in 3m42s, from 913 collected.

- The 2 skips are the Windows-only cache liveness paths (`tests/test_cache.py`),
  the same two that skip on this box on `main`. **No data-gated test skipped**;
  this ran in the main checkout, which has `data/`, so the ~118 tests that vanish
  from collection in a data-less worktree all ran.
- Before: 954 passed / 2 skipped of 956. After: 911 / 2 of 913. The delta is
  exactly the 43 removed graph tests, with no other test lost or newly skipped.
- `ruff check`, `ruff format --check`, and `mypy sisr` all clean.

**Real CUDA fit** — unit tests cannot cover the hooks this change removes, because
those hooks are the ones ordinary training runs through. Reference SRResNet config,
real DIV2K, 300 steps across three validation boundaries:

| val boundary | step | `psnr/val/RGB` |
| --- | --- | --- |
| 1 | 99 | 21.787 dB |
| 2 | 199 | 22.439 dB |
| 3 | 299 | 23.868 dB |

- All **225** float tensors differ between the step-99 and step-299 checkpoints;
  **0** unchanged. That is the direct refutation of the frozen-parameter failure
  the deleted hooks guarded against — a frozen run shows a moving loss curve, so
  loss alone would not have proven this.
- Checkpoints, bare weights, benchmark scoring and the grad-norm diagnostic all
  wrote as expected, and filenames stamp the batch axis (99/199/299).

## Checklist

- [x] Commit subjects use a Conventional Commit type and describe the effect
- [ ] Breaking changes carry `!` and a `BREAKING CHANGE:` footer — n/a, nothing released

## Merge strategy

> **Bottom of stack #160** (`#152 -> #153 -> #154 -> #155 -> #156 -> #159`). Merge this **first**.
> The PRs above it are linked into a GitHub stack and each runs full CI in its own right,
> so "green" above is real, not deferred.


YeapJieShen (commented):
Human reviewed, all good.

PR #153, opened 2026-08-26
refactor: group the metrics and the torch-free helpers into packages

### LINKED ISSUE #114: Reorganise modules into metrics/ and utils/ packages
## Proposal

Group the root-level modules into two packages:

- **`sisr/metrics/`** ← `ssim.py`, `perceptual.py`, and `scoring.py` (currently under `training/`,
  where it has nothing to do with training). `scoring.py` owns *"score an aligned `(sr_rgb, hr_rgb)`
  pair under this eval config"* — it is the orchestrator over exactly those two primitives, so the
  three together form one deep module rather than a folder.
- **`sisr/utils/`** ← `cache.py`, `imresize.py`.

`colorspace.py` stays out of `utils/`. It is not a utility: the full-range BT.601 variant is a
*training* colorspace — changing it renumbers a trained model's inputs — while the studio-range
variant is metric-only. That is a domain concept with paper-fidelity consequences, and filing it
under `utils/` hides it.

## Constraint that must not be broken

`sisr/cache.py` is **deliberately torch-free**, and its module docstring says so explicitly:
colorspace math was moved out of it for exactly this reason. There is a guard test asserting
`import sisr.cache` does not pull torch into `sys.modules`, because every spawned LMDB build
worker would otherwise pay a ~10 s torch import for nothing.

`colorspace.py` imports torch eagerly at module top level. So:

- `sisr/utils/__init__.py` **must not re-export** its submodules, or the guard breaks the moment
  anything torch-dependent joins the package.
- The guard test should be **extended to the package**, not left pointing at the module. The
  package boundary is what would silently rot.

This is invisible on a `fork` platform, where workers inherit already-imported pages. It is real
on `spawn` platforms — Windows and macOS contributors, and the Windows CI legs.

## Acceptance

- [ ] Modules moved, imports updated, public re-export smoke test green
- [ ] Torch-free guard extended to `sisr/utils/` as a package and proven to fail if violated
- [ ] Suite green

YeapJieShen:
Closes #114.

Stacked on #152 — **merge that one first.**

## What changed

```
sisr/metrics/     ssim.py  perceptual.py  scoring.py   (scoring.py from sisr/training/)
sisr/utils/       cache.py  imresize.py
sisr/colorspace.py                                     (stays at the root, deliberately)
```

Tests move to mirror the layout: `tests/metrics/`, `tests/utils/`.

## Why

Five root-level modules gave no hint which layer they belonged to. `scoring.py`
in particular sat under `sisr/training/` while having nothing to do with training
— it answers *"score an aligned `(sr_rgb, hr_rgb)` pair under this eval config"*,
which is the orchestrator over exactly the two primitives it now sits beside.

**`colorspace.py` stays at the root on purpose.** The full-range BT.601 variant is
a *training* colorspace — changing it renumbers a trained model's inputs — while
the studio-range variant is metric-only. Filing a distinction with that much
paper-fidelity consequence under `utils/` would hide it.

## The constraint that shaped this

`sisr/utils/cache.py` must stay torch-free: `parallel_build` fans out over a
`ProcessPoolExecutor`, and on spawn platforms every worker re-imports the module
tree holding its callable. `colorspace.py` imports torch eagerly.

So **neither package re-exports**, and for `utils/` that is load-bearing rather
than stylistic — a re-export means importing one member imports them all, so a
single torch-dependent addition would silently put a multi-second import back into
every build worker. The guard is now parametrised over the **package** as well as
its modules, because the package boundary is the thing that would rot.

This is invisible on `fork` platforms and real on `spawn` ones — Windows and macOS
contributors, and the Windows CI legs.

## Evidence

**Suite** — `913 passed, 2 skipped` of 915 collected, 3m26s.

- The 2 skips are the Windows-only cache liveness paths, the same two that skip on
  `main`. **No data-gated test skipped.**
- 915 vs #152's 913 is exactly the +2 from parametrising the guard into three cases.
- `ruff check`, `ruff format --check`, `mypy sisr` all clean.

**The torch-free guard proven to fail when violated** — a guard that passes both
ways documents nothing. Adding a single `from ..colorspace import rgb_to_ycbcr`
re-export to `sisr/utils/__init__.py`:

```
FAILED tests/utils/test_cache.py::test_utils_package_imports_no_torch[sisr.utils]
FAILED tests/utils/test_cache.py::test_utils_package_imports_no_torch[sisr.utils.cache]
FAILED tests/utils/test_cache.py::test_utils_package_imports_no_torch[sisr.utils.imresize]
```

All three fail, including the two modules that are themselves still torch-free —
which is precisely the "import one member, import them all" mechanism the rule
exists to prevent. Reverted, green again.

**Touches `sisr/utils/imresize.py`** — byte-equality against the MATLAB reference
sets re-proven:

- **21/21 passed, 0 skipped** (both reference sets present).
- The file's only diff is two docstring path references; behaviour is unchanged by
  construction, and `git diff` against the pre-move blob confirms it.

**An import cycle was found and closed.** `metrics/scoring.py` referenced
`SREvalConfig`, and `sisr.training` imports back into `sisr.metrics`. Made
type-only, since every runtime use is attribute access on the instance. Proven real
— with a runtime import instead:

```
ImportError: cannot import name 'SRScorer' from partially initialized module
'sisr.metrics.scoring' (most likely due to a circular import)
```

`metrics/` now has no runtime dependency on `training/`, which is the honest layering.

**One real type error surfaced.** Moving `scoring.py` out of `sisr.training.*` took
it out of mypy's blanket override and type-checked it for the first time:
torchmetrics widens its SSIM return to `Tensor | tuple[Tensor, Tensor]`. Narrowed at
the call site with the reason stated, not suppressed.

**Two silent-skip bugs caught.** `test_ssim.py` and `test_imresize.py` resolved
shared fixture directories relative to their own file. One broke collection loudly;
the other two would have made data-gated tests *skip* rather than fail. Both anchored
explicitly now. Caught because this ran in a checkout that has `data/` — the ~118
data-gated tests never materialise in a worktree.

## Checklist

- [x] Commit subjects use a Conventional Commit type and describe the effect

## Merge strategy


YeapJieShen (commented):
Human reviewed, all good.

PR #154, opened 2026-08-26
fix: bound every wait on a cache build lock, not just the live-holder one

### LINKED ISSUE #128: Cache stale-lock hard cap misses the dead-pid branch
## The gap

The cache's stale-lock handling has a hard cap at three times the lock timeout. That cap **misses
the not-stale / dead-pid branch**.

## Why this one is split out from the follow-up batch

It has a **silent-hang shape**. The other items in that batch are hygiene; this one can cause a
build worker to wait rather than fail, which is the failure mode hardest to notice and hardest to
diagnose after the fact.

## Acceptance

- [ ] A test proving the hang on unfixed code, red before the fix and green after
- [ ] The dead-pid branch covered by the cap, or an explicit reason it should not be

YeapJieShen:
Closes #128.

Stacked on #153, which is stacked on #152 — **merge bottom-up: #152, then #153, then this.**

## What changed

The build lock's hard cap now bounds **every** wait path, not only the one where
the holder's pid is confirmed alive.

## Why

A sentinel whose pid is **unreadable** — truncated, or read mid-write — or whose
pid is **this process's own after recycling** takes neither the live-holder path
(that branch requires a readable pid that is not ours) nor the abandoned-lock
takeover path, because a still-running heartbeat keeps its mtime fresh so it never
looks stale.

The waiter then polls forever. No exception, no progress, and no log line after
the first — which is the failure shape hardest to notice while it happens and
hardest to attribute afterwards. That is why this was split out of the follow-up
batch: the rest of that batch is hygiene, and this one hangs a build worker.

## How it is fixed

The deadline is evaluated **once per pass** and applies to every branch, rather
than being repeated inside the one branch that had it. A new wait path added later
inherits the cap instead of silently missing it — which is exactly how this branch
came to miss it. The two messages differ only in what they can honestly say about
the holder.

`_LIVE_HOLDER_WAIT_MULTIPLE` is renamed `_LOCK_WAIT_MULTIPLE`: it no longer bounds
only live holders, and a constant whose name understates its scope is how the next
branch gets missed.

## Evidence

**Test proven RED before the fix, GREEN after** — the acceptance criterion.

Against the unfixed code:

```
_SpunForever: polled 61 times without raising TimeoutError
1 failed in 1.34s
```

After the fix: `1 passed`.

The test replaces `time.sleep` rather than shortening it, so an unbounded loop
fails fast here instead of hanging the suite, and the **counter** is what
distinguishes "raised late" from "never raised at all" — a plain `pytest.raises`
would report the same failure for a hang and for a slow raise.

> One note on process, since it changes what the evidence is worth: my first
> version of this test appeared to go red for the right reason but did not. Patching
> `sisr.utils.cache.time.sleep` rebinds `sleep` on the real `time` module, so the
> fake recursed into itself, never actually slept, and the wall clock never reached
> the deadline. The red above is from the corrected test, re-run against the unfixed
> code.

**Suite** — `914 passed, 2 skipped` of 916 collected, 3m22s. The 2 skips are the

## Checklist

- [x] Test confirmed **RED** against the unfixed code, **GREEN** after

## Merge strategy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,11 @@ notice.
   README.
 - **The LMDB cache could delete another process's in-progress build.** Deletion is now
   restricted to proven corruption; environmental failures raise instead.
+- **A cache build could wait on its lock forever.** The timeout only bounded waits on a
+  holder confirmed alive. A sentinel whose pid was unreadable, or was this process's own
+  after pid recycling, took neither that path nor the abandoned-lock takeover path while
+  a heartbeat kept it looking fresh — so the waiter polled with no error, no progress and
+  no further log line. Every wait path is now bounded by the same cap.
 - **SRCNN's paper weight-init std** corrected to the published `0.001`.
 - **`example_input_shape` pointed at the wrong size**, defeating compile warm-up and
   misreporting FLOPs.
--- a/sisr/utils/cache.py
+++ b/sisr/utils/cache.py
@@ -41,9 +41,12 @@ _CORRUPTION_ERRORS = (lmdb.CorruptedError, lmdb.InvalidError, lmdb.VersionMismat
 _HEARTBEAT_MIN_INTERVAL = 0.01
 _HEARTBEAT_MAX_INTERVAL = 30.0
 
-# How many multiples of lock_timeout to wait on a confirmed-live holder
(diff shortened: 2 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #156, opened 2026-08-26
chore: make type checking a gate and narrow what it exempts

### LINKED ISSUE #116: Adopt static type checking and ship py.typed
## Proposal

Adopt static type checking as a gate and ship `py.typed` so downstream users get the annotations.

## Why now

This is the code-health foundation for the architecture-breadth work that follows. Landing it
before several more architectures arrive is substantially cheaper than retrofitting it after —
and it pairs naturally with the module reorganisation, since both change the import surface and
doing them as one arc avoids two rounds of churn.

## Notes

The codebase already uses modern type hints (PEP 585/604) throughout, and `mypy` currently runs
clean on the package, so this is about making it a *gate* and a *published guarantee* rather than
about fixing a backlog of annotations.

`py.typed` is a promise to downstream users. It should ship only once the checker is enforced in
CI, not before.

## Acceptance

- [ ] Type checking enforced in CI on the package
- [ ] `py.typed` shipped and included in the built distribution
- [ ] A deliberate violation is proven to fail the gate

### LINKED ISSUE #132: CI gate visibility: badge and README line
Add a badge and a README line stating that the test suite gates merges. Cheap, and it is the first thing a visitor uses to judge whether the numbers below can be trusted.

YeapJieShen:
Closes #116. Closes #132.

Stacked on #155 → #154 → #153 → #152 — **merge bottom-up.**

## What was already true

Worth stating, because it changes what this PR is: `mypy` already ran in CI,
`sisr/py.typed` already existed, and a test already built a real wheel to prove the
marker ships. #132 was already satisfied verbatim — the `test` badge and the line
about it were both in the README.

## What was actually missing

**The gate was not a gate.** `typecheck` was not among the required status checks
— only `test`, `build` and `lint` were — so it could fail and a pull request would
still merge.

**The exemption was a glob.** `sisr.training.*` swallowed every module added to
that package regardless of whether it would check clean. `scoring.py` was exempt
purely because of where it used to live; #153 moved it, and it immediately started
checking and immediately found a real widening.

## What changed

- `typecheck` **added to the required status checks** on `main-protection`.
  Verified against the server afterwards: `test, build, lint, typecheck`,
  enforcement `active`.
- The exemption list **names five modules individually** instead of matching a
  glob, so `sisr.training.config`, `.metadata` and `.probe` are now checked, and a
  new exemption is a visible line in a diff. The five that remain
  (`cli`, `callbacks`, `datamodule`, `gan_module`, `lightning_module`) are the ones
  where Lightning's hook system and jsonargparse's subclass resolution genuinely
  fight a checker.
- One real annotation gap surfaced and was fixed: the package's lazy
  `__getattr__` resolver, previously hidden by the glob.
- **README** names all four gates and what each covers, and states that `py.typed`
  ships — closing #132 with something load-bearing rather than closing it empty.

## Evidence

**The newly-covered modules are genuinely checked, not silently skipped.** I
injected a deliberate error into each and confirmed the checker reports it:

| module | result |
| --- | --- |
| `sisr.metrics.scoring` | CHECKED |
| `sisr.training.config` | CHECKED |
| `sisr.training.metadata` | CHECKED |
| `sisr.training.probe` | CHECKED |

**The gate is proven to bite, permanently.** `tests/test_typing.py` runs the
checker under this project's own config against a deliberate violation and requires
the specific `return-value` error — not merely a non-zero exit, which a crash, a bad
path or an unreadable config would also produce. It also runs it against a *clean*
module, because an error that fires on everything is not a gate either. And it pins
the exemption list, failing if a glob reappears or the list changes without the test
changing with it.

**Suite** — `933 passed, 2 skipped` of 935 collected, 3m21s. The 2 skips are the

## Follow-up, filed separately

Removing the five remaining exemptions is real work with an unknown floor, so it is
its own issue rather than a silent omission here.

## Checklist

- [x] A deliberate violation is proven to fail the gate

## Merge strategy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@ notice.
 
 ### Changed
 
+- **Type checking is a required status check**, and the package ships a `py.typed`
+  marker, so the annotations are a checked promise to downstream users rather than an
+  internal convention. The modules exempted from it are now named one by one instead
+  of matched by a glob that silently swallowed every module added beside them.
 - **`LMDBCache.close()`** releases the read environment and lets the next access reopen
   it. Closing the object `get_env()` returns left the instance holding a dead handle.
 - **Y-channel metrics are computed in BT.601 studio range**, matching MATLAB and the
--- a/README.md
+++ b/README.md
@@ -4,11 +4,16 @@
 [![build](https://github.com/YeapJieShen/single-image-super-resolution/actions/workflows/build.yml/badge.svg)](https://github.com/YeapJieShen/single-image-super-resolution/actions/workflows/build.yml)
 [![coverage](https://codecov.io/gh/YeapJieShen/single-image-super-resolution/branch/main/graph/badge.svg)](https://app.codecov.io/gh/YeapJieShen/single-image-super-resolution)
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
+[![mypy](https://img.shields.io/badge/mypy-checked-blue)](https://mypy-lang.org/)
 [![Python 3.12+](https://img.shields.io/badge/python-3.12%20%7C%203.13-blue)](https://www.python.org/downloads/)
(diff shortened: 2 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #159, opened 2026-08-26
docs: give the config surface a reference page and cut the templates back to config

### LINKED ISSUE #158: Move config rationale into a reference doc and cut the comment blocks back
## The problem

The tracked config surface is roughly half prose. On the current tree:

| file | lines | comment | share |
| --- | ---: | ---: | ---: |
| `templates/config.srgan.template.yaml` | 294 | 148 | 50% |
| `pyproject.toml` | 238 | 108 | 45% |
| `templates/config.srresnet.template.yaml` | 194 | 72 | 37% |
| `templates/config.srcnn.template.yaml` | 164 | 43 | 26% |
| `.github/workflows/test.yml` | 200 | 45 | 22% |
| `.github/workflows/build.yml` | 112 | 17 | 15% |

433 comment lines. A reader opening a template to see how an experiment is
configured has to read past measured throughput figures, dated platform results
and decision history to find the settings.

Two separate objections:

1. **Duplication.** A measured number, a date or a platform result has exactly one
   owner. Restating it beside a config key creates a second copy that ages
   independently — which is how a template ends up quoting a figure for code that
   no longer exists.
2. **Readability.** Config that is half changelog is hard to read as config.

`sisr/` itself is not in scope: 3.6% comment density, and only a handful of lines
cite a measurement.

## The rule

**One line per comment block, maximum.** Anything longer either moves to the
reference doc below, moves to wherever the measurement is actually owned, or goes.

Keep what tells a reader how to *use* the setting in front of them. Drop what
records how it was decided.

## What must survive the sweep

Not everything long is noise. These are load-bearing and must land somewhere, not
just be deleted:

- The `torch>=2.6` floor's CVE rationale — it is a security constraint, not trivia.
- bf16-mixed alone being a **regression** at this batch size, which saves someone a
  wasted experiment.
- Notes on settings that are mutually exclusive or that must be set together.
- The reason the `typecheck` job runs on `windows-latest` — it looks arbitrary and
  is not.

## Deliverable

A tracked `docs/configuration.md` owning the config surface, with the templates
reduced to near-bare YAML pointing at it. `.gitignore` needs no change: it ignores
`docs/superpowers/` and `/docs/agents/` specifically, not `docs/` wholesale.

## Acceptance

- [ ] No comment block longer than one line in the files above
- [ ] Every surviving rationale listed above is present in the reference doc
- [ ] No measured figure, date or platform result restated in a tracked config file
- [ ] `docs/configuration.md` covers every setting a template exposes
- [ ] Templates still parse and still train

YeapJieShen:
Closes #158.

Top of the stack: #156 → #155 → #154 → #153 → #152 — **merge bottom-up.**

## What changed

The tracked config surface was roughly half prose. `docs/configuration.md` now owns
the explanations, and every comment block in those files is one line.

| file | comment lines before | after |
| --- | ---: | ---: |
| `templates/config.srgan.template.yaml` | 148 | 24 |
| `pyproject.toml` | 108 | 22 |
| `templates/config.srresnet.template.yaml` | 72 | 13 |
| `templates/config.srcnn.template.yaml` | 43 | 11 |
| `.github/workflows/test.yml` | 45 | 8 |
| `.github/workflows/build.yml` | 17 | 7 |
| **total** | **433** | **78** |

## Why

Two separate problems, and only one of them is aesthetic.

**A measured number has exactly one owner.** Restating a throughput figure or a
platform result beside a config key creates a second copy that ages independently.
That is not hypothetical here: until #152 the SRResNet template's headline
throughput row described the graph-capture path, which by then was being deleted.

**Config that is half changelog is hard to read as config.** Someone opening a
template to see how an experiment is wired had to read past decision history to
find the settings.

`sisr/` is deliberately out of scope — 3.6% comment density, and only a handful of
lines cite a measurement.

## What survived, and where it went

Not everything long was noise. These moved rather than went:

- The `torch>=2.6` floor's **CVE rationale** — a security constraint, kept as a
  one-line `SECURITY FLOOR` marker in `pyproject.toml` and explained in full in the doc.
- **bf16-mixed is a regression, not a speed-up** at these batch sizes, and only wins
  paired with inductor — the thing most likely to be rediscovered the expensive way.
- The **global-step vs batch** units under SRGAN, including that an even
  `every_n_train_steps` halves the batch cadence while an odd one does not.
- Why **`typecheck` runs on `windows-latest`**, which looks arbitrary and is not.
- Settings only valid **together** (`precision` + `compile_backend`) or **anchored**
  to each other (`hr_input_size` ↔ `hr_crop_size`).
- Why **`GradNormLogger` is absent** from the SRGAN template.

## The templates are now strictly ASCII — they were not before

Worth flagging, since the project treats this as a hard rule: jsonargparse opens
config files with the platform encoding, so anything outside it is a launch-time
`UnicodeDecodeError`. All three templates already contained non-ASCII (em dashes, a
section sign) — 6, 7 and 12 lines. Those survived only because cp1252 happens to
decode them; they would not have survived a stricter platform encoding, and an emoji
in the same position would have failed outright. All three are pure ASCII now.

## Evidence

**Behaviour unchanged by construction, and checked** — each file parsed before and
after and the structures compared:

| file | result |
| --- | --- |
| all three templates | parse identical to their previous versions |
| `pyproject.toml` | parses identical |
| both workflows | parse identical, jobs and per-job permissions intact |

**All three templates still resolve through the CLI** (`--print_config`).

**The stripped SRResNet template still trains** — a real fit, 120 steps across two
validation boundaries:

- `psnr/val/RGB` 19.296 → 22.025 dB
- `loss/train` 0.1064 → 0.0392
- 4 checkpoints written

**Suite** — `933 passed, 2 skipped` of 935 collected, 3m31s. The 2 skips are the

**Every markdown link checked**, including the new doc's `../`-relative ones.

## Checklist

- [x] Doc-only change, in a single `docs:` commit

## Merge strategy

- [x] **Squash** — single-purpose, one `docs:` commit

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,7 @@ on:
     branches: [main]
   workflow_dispatch:
 
-# Workflow-wide floor. actions: write is NOT here -- it is granted per job, to
-# the ones that actually save an Actions cache, so a job that only checks out
-# and runs a script cannot write to the cache at all.
+# Floor only: actions: write is granted per job, to the ones that save a cache.
 permissions:
   contents: read
 
@@ -18,9 +16,7 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  # Cheap gate: skips the expensive matrices below on a doc-only PR. Deliberately
-  # an allowlist, not a denylist — anything not explicitly listed here (including
-  # this workflow file itself) is treated as code and runs the full matrix.
+  # Allowlist, not denylist: anything unlisted is treated as code. See docs/configuration.md.
   changes:
     runs-on: ubuntu-latest
     timeout-minutes: 5
@@ -66,18 +62,12 @@ jobs:
         run: pip install .
       - name: Smoke import
         run: python -c "import sisr; from sisr.cli import main; print('sisr', sisr.__version__, '- main:', main)"
-      # Ports the old tests/test_cli.py::test_cli_help_lists_subcommands subprocess
-      # smoke here: it cost ~10s of the pytest suite re-importing torch +
-      # lightning just to print --help, and this job already installs the wheel and
-      # smoke-imports it, so it's the natural home for a packaging-level check.
+      # Lives here, not in pytest: this job already has the wheel installed.
       - name: CLI entry point lists subcommands
         run: |
           python -c "import subprocess; out = subprocess.run(['sisr', '--help'], capture_output=True, text=True, check=True).stdout; missing = [s for s in ('fit', 'validate', 'test', 'predict') if s not in out]; assert not missing, f'sisr --help missing subcommands: {missing}'"
 
-  # Cheap single-leg guard for the *editable* install path (PEP 660): test.yml
-  # moved to uv, so nothing else in CI exercises `pip install -e`, yet
-  # CONTRIBUTING.md gives contributors who skip uv exactly that command. One
-  # OS/Python leg is enough — this checks build-backend behaviour, not the matrix.
+  # The only leg exercising `pip install -e`, which CONTRIBUTING.md still gives.
   editable-install:
     needs: changes
     if: needs.changes.outputs.code == 'true'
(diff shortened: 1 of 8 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #168, opened 2026-08-28
refactor: put the checkpoint monitor rule where its state already lives

### LINKED ISSUE #167: Fold the checkpoint monitor validator into the base class that replaced its reason
## Observation

The checkpoint monitor validation lives in a module-level function with **one caller**, which passes
its own class name in as an argument:

```python
_validate_monitor_metric(type(self).__name__, self.monitor, pl_module, mode=self.mode)
```

Its own docstring says it is "kept as a module-level function only because the rule is pure" and that
its arguments "are all `self` there".

## Why it exists, and why that reason is gone

The two checkpoint callbacks were siblings with no shared parent, so the rule could not be reached
through `super()`. They gained a shared base class, and nobody removed the function. A docstring on
the second callback still asserts the dead reason — that the classes "are siblings, not parent/child,
so the check can't be reused via `super()`" — which has been false since that base class landed.

## Deletion test

Folding it into the base-class method that already delegates to it makes the complexity vanish rather
than relocate. The tests already call the method, never the free function, so they do not change.

## Acceptance

- [ ] The free function is gone and the rule lives in the method
- [ ] The stale docstring claim is corrected
- [ ] No test changes were required

YeapJieShen:
Closes #167.

## What

`_validate_monitor_metric` was a module-level function with **one caller**, which passed its own
class name in as an argument:

```python
_validate_monitor_metric(type(self).__name__, self.monitor, pl_module, mode=self.mode)
```

Its own docstring conceded the point — "kept as a module-level function only because the rule is
pure", and its arguments "are all `self` there".

## Why it existed, and why that reason had gone

The two checkpoint callbacks were siblings with no shared parent, so the rule could not be reached
through `super()`. They gained a shared base class in #105 and nobody moved it. A docstring on
`SRWeightsCheckpoint.setup` still claimed the two classes "are siblings, not parent/child, so the
check can't be reused via `super()`" — false since that base class landed, and the evidence the
function had outlived its reason. That claim is corrected here too.

## Deletion test

Complexity vanishes rather than relocating. The interface loses a function; the rule reads
`self.monitor`/`self.mode` directly instead of being handed them. Net -22 lines.

## Verification

- **No test changed.** The tests already called `_validate_monitor`, never the free function —
  `git status` shows one modified file.
- Full suite on this branch: **933 passed, 2 skipped** (3m39s), identical to `main`'s baseline; the
  two skips are the usual Windows-only cache liveness paths.
- `ruff check`, `ruff format --check`, `mypy sisr` all clean.
- Both error messages are unchanged, so anything matching on their text still matches.

## Risk

LOW. No seam, no public contract, not `SRModel`/`SRProcessor`, not the metric path.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -593,68 +593,6 @@ class WeightHistogramLogger(Callback):
                 experiment.add_histogram(tb_name, param, global_step=step)
 
 
-def _validate_monitor_metric(
-    label: str, monitor: str | None, pl_module: lightning.LightningModule, mode: str
-) -> None:
-    """Raise if ``monitor`` names no logged metric, or is monitored in the wrong direction.
-
-    Implementation behind :meth:`_SRCheckpointBase._validate_monitor`, which is
-    how every caller reaches it — ``label``/``monitor``/``mode`` are all ``self``
-    there. Kept as a module-level function only because the rule is pure.
-
-    The direction half exists because both classes default to ``mode='max'``,
-    which is right for PSNR/SSIM and exactly inverted for LPIPS/DISTS — the
-    failure is silent, keeping the worst checkpoint of the run with nothing in
-    the logs or filenames to indicate it. Which perceptual metrics are
-    lower-is-better comes from ``PERCEPTUAL_METRICS`` rather than being
-    restated here, so a future higher-is-better perceptual metric would still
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

Issue #118, opened 2026-08-26
Multi-GPU (DDP) metric path is unverified


YeapJieShen:
## Context

Templates pin `devices: [0]`. Multi-GPU has never been exercised.

## The specific gap

`BenchmarkImageLogger` buffers per-rank and means locally. Under DDP the per-image tags and the
means would need rank guards or `sync_dist`, or the reported metrics will be per-rank values
presented as global ones.

## Status

YAGNI while single-GPU, and latent the moment anyone sets `devices: 2`. Filed so that it is a
known gap rather than a surprise.

## Acceptance

- [ ] Either rank-correct the metric path and prove it on two devices, or refuse `devices > 1`
      explicitly with a clear message — refusing is an acceptable resolution

YeapJieShen:
## Direction chosen: support, not refuse

This issue's acceptance offered two routes — rank-correct the metric path, or refuse `devices > 1`
explicitly. **The maintainer has chosen support.** The refusal option is closed.

## The gap is wider than this issue states

Verified `Snapshot @ 8c56324`. This issue names only the benchmark logger's per-process buffering.
Three more, all silent:

1. **`sync_dist` appears zero times** in the package and its tests. *Every* validation metric would
   be one process's shard mean presented as a global one — and the checkpoint monitors read those,
   so best-checkpoint selection would diverge per process.
2. **The bare-weights checkpoint writes from every process.** Its `_save_checkpoint` calls
   `torch.save` unconditionally. Its sibling is safe only by inheritance: Lightning's
   `Strategy.save_checkpoint` gates on `if self.is_global_zero`, and this override replaces exactly
   that method. The `is_global_zero` a few lines below it guards the logger notification, not the write.
3. **The benchmark sets would be split across processes.** The datamodule returns the benchmark
   loaders alongside the primary validation loader, and Lightning's distributed sampler is on by
   default — so a five-image benchmark set becomes a one- or two-image mean per process.

Only the adversarial module declares the single-device assumption today; the base module and every
callback inherit none of it.

## Revised acceptance

Verification is possible on a single-GPU machine over a CPU backend — see the harness issue this is
blocked by. That bounds what can be claimed:

- [ ] Metrics reduce to a global value, proven on two processes
- [ ] The weights file is written by one process only, proven on two processes
- [ ] Benchmark sets are not split across processes
- [ ] The single-device assumption is stated at one interface, not per subclass
[...]

PR #169, opened 2026-08-28
test: make distributed code paths provable on a single-GPU machine

### LINKED ISSUE #166: A two-process test harness, so distributed code paths can be proven on one machine
## Context

No test in the suite runs more than one process, so nothing exercises a distributed code path. Any
rank-sensitive change therefore ships unproven, which conflicts with the standing rule that a
correctness fix arrives with a test that was red before it.

## What this is

A test harness that runs a real two-process trainer using the gloo backend on CPU:

```python
Trainer(accelerator="cpu", devices=2, strategy="ddp")
```

`gloo` is available in this environment (checked), and the suite already runs CPU-backed integration
tests, so the pattern exists.

## What it can prove

- A logged metric reduces to a global value rather than one process's shard
- A file is written by one process rather than all of them
- A dataset is or is not split across processes
- A collective called by one process and not another fails, rather than hanging silently in production

## What it cannot prove

NCCL-specific behaviour, device placement on real GPUs, memory, or multi-GPU throughput. Those need
actual multi-GPU hardware. Containers do not help — they share the host's devices, so two containers
on one GPU is still one GPU.

## Why it comes first

It is the enabler for every rank-sensitive change in this arc. With it in place, a rank guard can be
proven at the moment it is written, instead of being written once and proven later.

## Acceptance

- [ ] A reusable fixture runs a two-process trainer on CPU
- [ ] It is stable across repeated runs, not green once

YeapJieShen:
Closes #166. Second in the stack, on top of #168.

## Why it comes first

Nothing in the suite ran more than one process, so every rank-sensitive change in this arc would
have shipped unproven — which the standing fix criteria do not allow. With this in place, a rank
guard can be proven *at the moment it is written*.

## What it does

A fixture starting a real two-process trainer over **gloo on CPU**. No second GPU involved.

`ddp_spawn` rather than `ddp`: `ddp` re-executes the entry script, which under pytest means
re-running the whole test command. The launcher differs from production; the semantics under test —
sampler splitting, metric reduction, rank identity — belong to the strategy and are identical.

## The mechanism test asserts both directions

```
probe/reduced    = 0.5   # sync_dist=True  -> mean over both processes
probe/unreduced  = 0.0   # sync_dist=False -> the reporting process's own value
```

The second line is the live defect shape for validation metrics (#118), so it is asserted rather
than described.

## What it cannot prove

NCCL behaviour, device placement on real GPUs, throughput. Those need genuine multi-GPU hardware —
containers do not substitute, since they share the host's devices.

## Verification

- **Stable across five consecutive runs**, per the convention that one green run is not evidence
  for a new test.
- Costs ~23s of suite time.
- Test-only: no `sisr/` code changed.

## Risk

LOW. Touches no seam, no contract, no production code.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,9 @@
 """Shared pytest fixtures for the sisr test suite."""
 
+from collections.abc import Callable
 from pathlib import Path
 
+import lightning
 import numpy as np
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #170, opened 2026-08-28
fix: refuse to write an artifact that cannot state its own scale

### LINKED ISSUE #164: SRCNN artifacts record no scale, so they cannot say what they upscale by
## Observation

Built exactly as `templates/config.srcnn.template.yaml` builds it, an SRCNN artifact records **no
scale**:

```
Snapshot @ 8c56324
SRCNN     io: {'scale': None, 'input': 'pre_upsampled', 'input_channels': 1,
               'output_range': [0.0, 1.0], 'output_colorspace': 'Y'}
SRResNet  io: {'scale': 4, 'input': 'native_lr', ...}
```

## Why it happens

The metadata builder reads the training config's `scale`, falling back to the model's own
hyperparameters. SRCNN is `pre_upsampled` — the low-resolution input arrives already at output size,
so scale is a property of the **dataset**, not the model, and SRCNN has no `scale` hyperparameter to
fall back on. The training config field exists and no shipped template sets it.

Related: the same root cause makes the YAML anchor for scale live on the *model* for one
architecture family and on the *dataset* for the other.

## Why it matters

For a pre-upsampled model, the scale factor is the one thing a consumer must know — they have to
upsample the input by that factor before feeding it. The file cannot say whether it is x2, x3 or x4.
It also blocks the artifact-filename work, whose scale token has nothing to read.

## Acceptance

- [ ] The shipped SRCNN template records a scale
- [ ] Writing provenance with no resolvable scale fails loudly rather than recording null, so this
      cannot recur silently for a future pre-upsampled architecture
- [ ] A test proves the null case is refused

YeapJieShen:
Closes #164. Third in the stack.

## The defect

Built exactly as the shipped SRCNN template builds it, an artifact recorded no scale:

```
Snapshot @ 8c56324
SRCNN     io: {'scale': None, 'input': 'pre_upsampled', ...}
SRResNet  io: {'scale': 4,    'input': 'native_lr',     ...}
```

SRCNN is resolution-preserving, so scale is a property of the **data**, not the network — there is
no hyperparameter to fall back on. `SRTrainingConfig.scale` exists and no template set it.

## This changes a documented position, deliberately

The field's docstring said `None` "leaves it unchecked and out of exported metadata — **appropriate**
for architectures like SRCNN". That reasoning held while we were the only readers. It stops holding
for a file handed to someone else: for a pre-upsampled architecture the scale is precisely what they
must resize the input by, so a file that cannot state it is unusable. It also leaves any
artifact-naming scheme with nothing to read.

**What is preserved:** the scale is still *not* inferred from the datamodule. That refusal had a real
reason — `scale` is outside the `SRDataset` contract and `PredictDataset` has none — and it stands.
It is stated in the config, which is where per-paper knobs already live.

## Side effect worth having

The template's YAML anchor moves into the model block, which settles an inconsistency: the other two
templates already anchor scale there, and SRCNN anchored it on the dataset only because it had
nowhere else to put it.

## Verification

- RED first: `DID NOT RAISE ValueError` on pre-fix code.
- ~30 test modules were building artifacts with an unresolvable scale and now declare one.
- **Two tests whose subject *is* the scale-absent case keep it** — I reverted my own over-broad edit
  on both.
- One test asserted the removed behaviour directly and is replaced by its inverse.
- Full suite: 935 passed, 2 skipped.

## Also fixed in passing

A latent order dependence: a CLI test config had no seed, so the "No seed found" warning — fatal
under the strict filter — fired only when no earlier test had seeded the process. It **passed in a
full run and failed alone**.

## Risk

Not LOW. Refusing a previously-accepted value is a contract change.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -104,7 +104,19 @@ to the torch hub cache, and TV's `2e-8` presumes the `[-1, 1]` output range that
 | `compile_backend` | A `torch._dynamo` backend name, or unset for eager. |
 | `layer_lrs` | Per-`Conv2d` learning rates; requires every trainable parameter to live in a `Conv2d`. |
 | `init_strategy`, `init_mean`, `init_std` | Weight initialisation. |
-| `scale` | Upscaling factor, for provenance and cross-checks. |
+| `scale` | Upscaling factor. **Required whenever a run writes an artifact and the model has no `scale` of its own.** |
+
+`scale` is stated here for SRCNN and inherited from the network for SRResNet, and that
+asymmetry is real rather than an inconsistency to tidy away. SRResNet upsamples internally, so
+its scale is a property of the network and lives in its hyperparameters. SRCNN is
+resolution-preserving — it refines an input that has *already* been upsampled — so its scale is
+a property of the **data**, and nothing about the network records it.
+
+Set it anyway. Saving an artifact whose scale cannot be resolved from either source is refused
+rather than recorded as null: a file that cannot state its factor is unusable by anyone who did
+not train it, because for a pre-upsampled architecture the factor is exactly what they must
+resize the input by before feeding it. It is deliberately *not* inferred from the datamodule —
+`scale` is not part of the dataset contract, and a predict-only dataset has none at all.
 
 `example_input_shape` is **not** only a TensorBoard-graph dummy. It also shapes the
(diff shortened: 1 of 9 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #171, opened 2026-08-28
feat!: give the distributable artifact an owner, and make it safetensors

### LINKED ISSUE #125: Checkpoint provenance has no reader-side owner
## Observation

Checkpoint provenance metadata is *built* by a dedicated module, but nothing owns *reading* it.

About 90 lines of read-side validation live inside the adversarial module's generator-initialisation
path, reachable only by that one paradigm, while the export path loads a checkpoint with **no**
provenance check at all.

## Position: held, not proposed

One adapter is a hypothetical seam. Extracting a reader now would be designing against a single
caller, which is how shallow modules get built.

## Trigger

Revisit when a **second** reader appears — most likely when the export path or a resume path needs
the same validation the adversarial module already performs.

### LINKED ISSUE #162: Distributable weights move to safetensors, and .pt is removed
## Context

The distributable weights file is written as a `torch.save` payload of `{state_dict, meta}`. Nothing
in the repo writes `safetensors`, and the format is decided at the one place that writes it.

## What changes

- `safetensors` becomes the only format for the **distributable** artifact. `.pt` is removed, writer
  and reader — a clean break, not a deprecation.
- `safetensors` joins the **core** dependencies. It cannot be optional if it is the default save format.
- The provenance block maps onto the safetensors header as **one entry per top-level field**, not a
  single JSON blob. This matches what the ONNX sink already does, and for the same reason: a
  per-field table is readable by whatever inspects the file.
- `FORMAT_VERSION` moves to `v2`. The envelope's carrier and encoding both change and `.pt` stops
  existing, so this is a change to the existing shape rather than a new artifact kind.
- The ONNX sink's metadata is aligned field-for-field with it, so one description has two carriers.

## Explicitly out of scope

The resumable `.ckpt`. Lightning owns that format, and it carries optimizer state, loop state and
hyperparameters, none of which flat tensors can represent. **This does not make the repo
pickle-free** — it makes the file we hand to strangers pickle-free, which is the one that matters.

## Why this is a safety improvement, stated precisely

Every `torch.load` in this repo already passes `weights_only=True`, and the torch floor is pinned
with a CVE note, so *our* readers were never the exposure. The exposure is a **consumer** loading a
published `.pt` with `weights_only=False` or an older torch. `safetensors` removes that by
construction rather than by mitigation.

## Acceptance

- [ ] Distributable weights are written as `.safetensors`; no code path writes `.pt`
- [ ] Provenance survives a write/read round-trip, per field
- [ ] The three readers go through one module rather than re-deriving the payload shape
- [ ] ONNX metadata and the safetensors header carry the same fields

YeapJieShen:
Closes #162. Closes #125. Fourth in the stack.

## What was wrong

The bare-weights file was written in one place as a `torch.save` payload, and its shape re-derived
by hand in **three** readers — one of which, the export path, checked nothing at all. The format was
decided at the write site, so a second format meant editing four files and the readers had no way to
agree with each other.

`sisr/artifacts.py` now owns writing, reading and checking it. **Top-level, not under `training/`**:
reading a published artifact is not a training concern, and two of its three callers do not train.
That placement is what closes #125, whose held trigger — "revisit when a second reader appears" —
fired when a second *format* arrived.

## The safety argument, stated precisely

Not the obvious one. Our own readers were never exposed: every `torch.load` here passes
`weights_only=True` against a pinned torch floor. The exposure is a **consumer** opening a published
file with that turned off, or on an older torch. safetensors removes it by construction.

**This makes the file we hand out pickle-free, not the repository.** The resumable `.ckpt` is still a
pickle and stays one — Lightning owns that format, and it carries optimizer moments, loop state and
hyperparameters that a flat tensor map cannot represent.

`safetensors` is a **core** dependency, not an extra: it is what every run writes, so it cannot be
conditional on an optional install.

## Header encoding

One entry per top-level field, not one blob — the call the ONNX sink already made, for the same
reason. That sink now **shares this encoder** rather than keeping its own copy.

`FORMAT_VERSION` → `v2`. Carrier and encoding both changed and a format stopped existing, which is a
change to the existing shape rather than a new kind (the precedent for *not* bumping was a new kind).

## Validation splits by consequence

| | |
|---|---|
| **Refuse** | processor, output range — being wrong yields a plausible image and no error |
| **Warn** | library versions — cannot change what the tensors mean, and refusing would expire every artifact on the next dependency bump |
| **Refuse, stricter** | generator init keeps its five-field list — the wrong weights there silently *mistrain* |

## The rank gate

`ModelCheckpoint` gets this free by writing through `Strategy.save_checkpoint`, which gates on
`is_global_zero`. This callback overrides **exactly that method**, so every process was racing to
write the same path. Bookkeeping stays unconditional as in the base, and the rolling window's deletes
were already gated inside `Strategy.remove_checkpoint` — verified, not assumed.

Proven both ways: removing the guard fails the `is_global_zero=False` case, keeping it passes both.

## Verification

- 10 new tests for the artifact contract; 2 for the rank gate.
- Full suite: **947 passed, 2 skipped**. `ruff`, `ruff format`, `mypy` clean.

## Also corrected

`docs/configuration.md` still described checkpoint filenames as carrying optimizer steps, with a
worked example claiming a file named 10000 is the state at TensorBoard x=5000. **True before #111,
wrong since.** Found while documenting the format.

## BREAKING CHANGE

Bare-weights artifacts are `.safetensors`; existing `.pt` files cannot be read. No conversion path in
the tree, by decision — regenerate by retraining.

## Risk

Not LOW. Public artifact contract, and a breaking format change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,8 @@ sisr fit --config templates/config.srgan.template.yaml
 
 **The generator starts from an MSE-trained SRResNet.** Ledig et al. scope the
 MSE-initialisation trick to "when training the actual GAN", so a paper-faithful run
-points `training_config.init_from` at a finished SRResNet run's bare-weights `.pt` —
+points `training_config.init_from` at a finished SRResNet run's bare-weights
+`.safetensors` —
 never the sibling `.ckpt`, which holds the whole LightningModule. That file's own
 provenance metadata is checked against this run's generator, processor, output range and
 scale, and refused on any mismatch: weights trained under a different one produce a model
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -243,13 +243,44 @@ So the units differ, and you must read this before changing any of them:
 | In **global steps** | In **batches** |
 | --- | --- |
 | `max_steps` | `val_check_interval` |
-| checkpoint filenames | `log_every_n_steps` |
-| `every_n_train_steps` | hand-stepped LR scheduler `milestones` |
+| `every_n_train_steps` | `log_every_n_steps` |
+| | checkpoint filenames and metadata |
+| | hand-stepped LR scheduler `milestones` |
 
 `_batches_that_stepped` is also the **default x-axis every logged metric is plotted
-against**. Under SRGAN, TensorBoard therefore shows the run in batches while checkpoint
-filenames carry global steps — `sr-weights-10000.pt` is the state at TensorBoard
-x=5000, not x=10000.
+against**, which is why checkpoints are stamped with it: a saved file exists to be
+located on a curve, and one named in optimizer steps cannot be. So
+`sr-weights-10000.safetensors` is the state at TensorBoard x=10000 under every
+paradigm. The artifact's own metadata records **both** counters under distinct names —
+`global_step` keeps meaning the optimizer count, `batch_step` is the axis above — so a
+reader that only knows the older field still reads a true value.
+
+Note the asymmetry in the table: `every_n_train_steps` is Lightning's own and counts
+optimizer steps, while the filename it produces counts batches. Under SRGAN that means
+a cadence of 10000 writes a file named for batch 5000.
+
+### The distributable artifact
+
+`SRWeightsCheckpoint` writes **safetensors**, and it is the only format it writes — the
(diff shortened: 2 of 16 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #173, opened 2026-08-28
feat!: name artifacts after the model they hold

### LINKED ISSUE #163: Artifact filenames should say what the model is
## Context

Artifact filenames record **when** a file was written and **how good** it scored, and never **what it
is**: `sr-weights-{step}.pt`, or `sr-weights-{step}-psnr_val_RGB={value}.pt`. A folder of weights
cannot be read without loading each file.

## What changes

Filenames state identity at a glance:

```
<arch>_x<scale>_<colorspace>_<variant>_s<step>.safetensors

SRCNN_x2_Y_915_s500000.safetensors
SRResNet_x4_RGB_16B64F_s500000.safetensors
SRDiscriminator_96_s500000.safetensors
```

- The **variant** token is architecture-specific — `915` is SRCNN's kernel sizes, `16B64F` is
  SRResNet's block and filter counts. No generic rule over hyperparameters produces both without
  becoming unreadable, so `SRModel` gains a short variant-tag member and each architecture supplies
  its own.
- A **discriminator** uses the same grammar minus scale and colourspace, neither of which describes a
  critic.
- `s<step>` stays on the batch axis, unchanged. Note this already matches "optimizer steps applied to
  *these* weights" at the shipped `d_steps_per_g_step: 1`; the trainer's own step counter counts
  steps by *any* optimizer and matches neither.
- No monitored-metric token. Different runs monitor different metrics and adversarial runs monitor
  none, so it would be inconsistent between files that otherwise look alike. The value stays in the
  provenance block.

## Risk

`SRModel` gains a member, so **every architecture must supply it**. This is a base-class contract
change and is not a low-risk edit.

## Blocked by

The SRCNN scale defect — the `x<scale>` token has nothing to read for a pre-upsampled architecture
until that is fixed.

## Acceptance

- [ ] Both architectures and the discriminator produce names matching the grammar
- [ ] Every architecture supplies a variant tag; a new one cannot omit it silently
- [ ] Two saves in one run do not collide

YeapJieShen:
Closes #163. Fifth in the stack.

## Before / after

```
sr-weights-500000-psnr_val_RGB=28.8635.pt     ->  SRResNet_x4_RGB_16B64F_s500000.safetensors
sr-weights-500000.pt                          ->  SRCNN_x2_Y_915_s500000.safetensors
d-weights-500000.pt                           ->  SRDiscriminator_96_s500000.safetensors
```

Architecture, scale, colourspace, variant, step. A directory of weights is now readable with `ls`.

## The design point

The name is **derived from the artifact's own provenance**, not assembled beside it. A filename and a
header that disagree is the defect class this project keeps rediscovering — deriving one from the
other makes it unrepresentable. `build_metadata` gained the variant, `artifacts.stem` projects a name
from it, and the callback applies that in `setup()` — the earliest point at which the module exists,
since a callback built from YAML has no model.

## `SRModel.variant_tag`

Abstract, **never defaulted**, for the reason `input_contract` is declared rather than inferred: no
rule over `hparams` yields a readable tag for every architecture, and an inherited default would
label two different configurations identically. SRCNN gives its kernel triple (`915`), SRResNet its
blocks and width (`16B64F`). The discriminator supplies one too, duck-typed, since it is deliberately
not an `SRModel`.

## Two deliberate absences

**No metric token.** Runs monitor different metrics and adversarial runs monitor none, so the token
made otherwise-identical files look unrelated. The value is in the header. As a *side effect* the
`/`-sanitisation hazard disappears rather than being handled — a monitor tag can no longer reach a
path at all.

**`s<step>` is the batch axis**, unchanged from #111, so the number can be found on a curve.

## One structural fix found on the way

Validation became a hook, so a subclass's checks run **before** the name is derived from the very
component they confirm exists. Without it, a bad `attribute` surfaced as a bare `AttributeError` from
inside the naming code instead of the actionable message that already existed.

## Test approach

Tests whose subject is the rolling window, the step axis or the payload now **pin an explicit
prefix** — they should not move every time naming does. New tests cover the identity end-to-end, the
override, and that a subclass omitting `variant_tag` cannot be instantiated.

Full suite: **955 passed, 2 skipped**. `ruff`, `ruff format`, `mypy` clean.

## BREAKING CHANGE

Filenames change shape, and every `SRModel` subclass must supply `variant_tag`.

## Risk

Not LOW — base-class contract change.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -276,6 +276,30 @@ Provenance travels in the file's header, one entry per top-level field rather th
 opaque blob, so anything that can open the artifact gets a readable table. The ONNX
 export writes the same fields the same way.
 
+Filenames say what the model is, so a directory of weights can be read without
+opening anything:
+
+```
+SRCNN_x2_Y_915_s500000.safetensors
+SRResNet_x4_RGB_16B64F_s500000.safetensors
+SRDiscriminator_96_s500000.safetensors
+```
+
+Architecture, scale, colourspace, variant, step. The **variant** is
+architecture-specific — SRCNN's kernel triple, SRResNet's block and filter counts — and
+each architecture supplies its own via `variant_tag`; no generic rule over
+hyperparameters produces a readable tag for all of them. A component drops scale and
+colourspace, since neither describes a critic.
+
+The name is **derived from the artifact's own provenance**, not assembled separately, so
+a file whose name and header disagree is not representable. Set `filename_prefix` to
+override it for a run that wants its own naming.
+
+Two things are deliberately absent. There is **no metric token**: runs monitor different
+metrics and adversarial runs monitor none, so it would make otherwise-identical files
+look unrelated — the value is in the header instead. And `s<step>` is the **batch** axis,
+which is the axis every logged metric uses, so the number can be found on a curve.
+
 Mismatches are checked on read. Fields that change what the output *means* — the
 processor and the output range — are refused, because being wrong about either produces
 a plausible image and no error. Library-version drift is warned about and loaded, since
--- a/sisr/artifacts.py
+++ b/sisr/artifacts.py
@@ -175,3 +175,39 @@ def require_compatible(
             UserWarning,
             stacklevel=2,
         )
+
+
+def stem(meta: dict[str, Any]) -> str:
+    """Filename identity for an artifact, derived from its own provenance.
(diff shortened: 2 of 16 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #174, opened 2026-08-28
refactor!: give config keys their unit, and stop restating a derived value

### LINKED ISSUE #165: Config keys should name their own unit, and stop restating derived values
## Context

Two usability problems in the config surface, both verifiable by reading the shipped templates.

### One name, two meanings

`log_every_n_steps` appears twice in every template. On the trainer it is the metric **flush**
cadence; on our own diagnostic callback it is a **sampling** cadence. Same name, same file, different
meaning. The flush cadence already has a documented failure mode of its own — it quantises any
throughput figure read back from the logged metrics.

Lightning's own callbacks name the unit in the parameter (`every_n_train_steps`, `every_n_epochs`).
Our callback is the deviation: it copied a **trainer** argument name onto a **callback**. Renaming
ours to `every_n_batches` follows that convention rather than breaking it, and stops the collision.

The trainer's own names stay untouched. The residual ambiguity there — one counter is optimizer
steps, two others are batches — is Lightning's, and the honest fix is to **state the conversion at
run start** when the two axes differ, not to invent names.

### Arithmetic the config does by hand

`example_input_shape` is written out as a literal that must equal a value the code already derives
and already validates:

- one family: crop size divided by scale
- the other: the sub-image size, with no division

Two different derivation rules for one field, both computed by the reader. A value the code can
check is a value the code can derive. Making it optional keeps the ability to state an intent and
have it verified, while the templates stop carrying arithmetic.

## Not in scope

Splitting configs into a base plus overrides. That was considered and rejected — one self-contained
file per experiment is a settled decision.

## Acceptance

- [ ] Our cadence parameters name their unit and no longer collide with a trainer argument
- [ ] `example_input_shape` may be omitted and is derived; an explicit value still overrides and is
      still checked
- [ ] The templates lose the comments that existed only to disambiguate units

YeapJieShen:
Closes #165. Sixth in the stack.

## One name, two meanings

`log_every_n_steps` appeared **twice in every template**, meaning different things: the trainer's
metric *flush* cadence, and a diagnostic callback's *sampling* cadence.

Ours are now `every_n_batches` and `every_n_val_runs`. That **follows** Lightning's convention rather
than breaking it — its own callbacks already name the unit (`every_n_train_steps`, `every_n_epochs`),
and copying a *Trainer* argument name onto a *callback* was the deviation.

## The residual asymmetry, handled honestly

Lightning's own names cannot change, so `max_steps` and `every_n_train_steps` still count optimizer
steps while everything else counts batches. Documentation was not enough — a `max_steps` copied from
the SRResNet template trains an adversarial run **half as long**. So the run now *prints the
conversion at start*, against its own numbers:

```
SRGAN takes 2 optimizer steps per batch, so max_steps=400000 is 200000 batches.
val_check_interval, every_n_batches and the hand-stepped scheduler milestones all
count batches; max_steps and every_n_train_steps count optimizer steps.
```

Five capitalised warning comments come out of the SRGAN template as a result.

## Arithmetic the config did by hand

`example_input_shape` was written out under **two different rules** — crop size over scale for a
native-LR dataset, sub-image size for a pre-upsampled one — and then validated against the very
tensor those rules describe. A value the code can check is a value the code can supply.

It is now optional and derived. An explicit value still wins **and is still checked**, so a config can
state an intent and have it contradicted.

**The templates keep an explicit value**, and the reason is now stated where it is set: they are also
the input to `sisr export`, which never sees training data and has nothing to derive from. I removed
it first and the export subcommand broke — worth saying, because it is the one thing that made this
not a pure simplification.

## Not in scope, deliberately

Splitting configs into a base plus overrides. Considered and rejected before — one self-contained
file per experiment.

## Verification

Full suite: **957 passed, 2 skipped**. `ruff`, `ruff format`, `mypy` clean. New tests cover the
derivation and that an explicit value is still contradicted.

## Note on untracked configs

The gitignored golden run configs still use the old key names. Left alone on purpose — they record
what those runs actually used, and editing them would falsify that.

## BREAKING CHANGE

`GradNormLogger`/`WeightHistogramLogger` take `every_n_batches`; `BenchmarkImageLogger` takes
`every_n_val_runs`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -100,7 +100,7 @@ to the torch hub cache, and TV's `2e-8` presumes the `[-1, 1]` output range that
 
 | Field | Notes |
 | --- | --- |
-| `example_input_shape` | Must match the real training input. |
+| `example_input_shape` | **Optional.** Derived from the real training patch when unset; checked against it when set. |
 | `compile_backend` | A `torch._dynamo` backend name, or unset for eager. |
 | `layer_lrs` | Per-`Conv2d` learning rates; requires every trainable parameter to live in a `Conv2d`. |
 | `init_strategy`, `init_mean`, `init_std` | Weight initialisation. |
@@ -121,6 +121,12 @@ resize the input by before feeding it. It is deliberately *not* inferred from th
 `example_input_shape` is **not** only a TensorBoard-graph dummy. It also shapes the
 `torch.compile` warm-up — a wrong size compiles once and then recompiles on step 1 —
 and it drives `ModelSummary`'s FLOPs count.
+
+You may leave it unset: it is derived from the first real training patch. Setting it is
+still useful, because a stated value is *checked* against that patch rather than
+replaced by it — an intent that can be contradicted. **The shipped templates state it**
+for one reason: they are also the input to `sisr export`, which never sees training data
+and so has nothing to derive from.
 
 ### `eval_config` and the two SSIM conventions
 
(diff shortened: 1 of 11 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #175, opened 2026-08-28
feat: make the metric path correct under distributed training

### LINKED ISSUE #118: Multi-GPU (DDP) metric path is unverified

YeapJieShen:
Closes #118. Top of the stack.

## The state before

Only the adversarial module said anything about multi-GPU. The base module and every callback **failed
open**: nothing was wrong until someone set `devices: 2`, and then nothing failed either — the numbers
were just wrong.

## Three silent defects

**1. `sync_dist` appeared nowhere.** Every validation metric was one process's shard mean presented as
global, and the checkpoint monitors read those — so different processes would have kept different
checkpoints. All validation and benchmark metrics now reduce. Per-step *training* scalars deliberately
do not: they are already the process's own, and reducing one would stall every step.

**2. The evaluation loaders would have been split.** This is the subtle one, and it is why this PR
looks the way it does. `DistributedSampler` pads a set to a multiple of the world size by
**repeating** samples. A five-image benchmark across four processes reports a figure for *eight*
images with three counted twice — not that set's score, on the metric path this project's central
claim rests on.

So the datamodule shards the **training** loader itself and leaves evaluation whole on every process.
Lightning's injection is one trainer-wide switch, so a distributed run **refuses to start** with it
still enabled rather than silently re-splitting them.

**3. The bare-weights write happened on every process** — already fixed earlier in this stack.

## The padding is demonstrated, not asserted

The test found it rather than assuming it: three training samples across two processes produce **four**
batches, not three. That is the padding, visible. Harmless while training, and exactly why evaluation
must not go through it.

## Verification, and its limits

Proven on **two real processes over gloo, on CPU**, using the harness at the bottom of this stack — so
rank-correctness is demonstrated rather than reasoned about, on a single-GPU machine. Stable across
three consecutive runs.

**Not covered, and the docs say so:** NCCL behaviour, device placement on real GPUs, throughput. Those
need genuine multi-GPU hardware; containers do not substitute, since they share the host's devices.

## Scope note

`SRGANLightning` **still refuses**. Manual optimization opts out of Lightning's gradient
synchronisation, so the two networks' gradients would not be reduced and each process would train its
own divergent pair with nothing failing. Supporting that means writing the synchronisation explicitly
— its own piece of work, not a line in this one. The models are covered; the adversarial paradigm is
not.

Full suite: **960 passed, 2 skipped**. `ruff`, `ruff format`, `mypy` clean.

## Risk

Not LOW. Metric path.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -265,6 +265,43 @@ Note the asymmetry in the table: `every_n_train_steps` is Lightning's own and co
 optimizer steps, while the filename it produces counts batches. Under SRGAN that means
 a cadence of 10000 writes a file named for batch 5000.
 
+### Multi-GPU
+
+`SRLightning` runs distributed. Two settings are required together:
+
+```yaml
+trainer:
+  devices: 2
+  strategy: ddp
+  use_distributed_sampler: false     # required -- see below
+```
+
+**`use_distributed_sampler: false` is not optional, and a distributed run refuses to
+start without it.** This project shards the *training* loader itself and leaves the
+validation and benchmark loaders whole on every process. Lightning's injection is a
+single trainer-wide switch, so leaving it on would split those too — and
+`DistributedSampler` pads a set to a multiple of the world size by **repeating** samples.
+A five-image benchmark across four processes becomes eight, three counted twice, and the
+reported figure is no longer that set's score. Every process therefore evaluates every
+benchmark image; the cross-process reduction on those metrics is then exact, and doubles
+as a check that the processes have not diverged.
+
+Every validation metric is reduced across processes. Without that each process reports
+its own value as though it were global, and the checkpoint monitors read those and pick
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #177, opened 2026-08-29
docs: record the artifact arc in the changelog

YeapJieShen:
The seven-PR artifact arc landed with no changelog entry, and it carried **three**
`BREAKING CHANGE:` commits. The `SRWeightsCheckpoint` bullet also still described a `.pt`
file that no longer exists.

## What this adds

**Added** — `sisr/artifacts.py` as the one owner of the distributable artifact; DDP
correctness across the metric path (including why evaluation loaders are deliberately
never sharded); the two-process test harness.

**Changed** — the three breaks, marked as breaks:
- distributable weights are `.safetensors` and `.pt` is removed, with no converter
- artifact and checkpoint filenames name architecture/scale/colorspace/variant/step
- `log_every_n_steps` on our own callbacks became `every_n_batches` / `every_n_val_runs`

plus `example_input_shape` becoming optional.

**Fixed** — an SRCNN artifact recorded no scale, so it could not say what it upscales.

## On calling them breaking

Nothing is released, so strictly these break no version. The repository is public and
people track `main`, so the entry says so explicitly rather than folding the breaks
silently into a description of the current state.

Doc-only. No issue — this was owed by the merge itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,21 @@ notice.
 
 ### Added
 
+- **`sisr/artifacts.py`, one owner for the distributable weights artifact** — save, load,
+  metadata encoding, the compatibility check a reader runs before trusting a file, and the
+  filename stem. The checkpoint callback, `sisr export` and the CLI all go through it, so
+  the three cannot disagree about what an artifact is. It sits at the top level rather than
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #178, opened 2026-08-29
ci: stop a cancelled run publishing a red required check

### LINKED ISSUE #161: Cancelled workflow runs publish a red required check and block merges
## The defect

Both workflows end in a gate job whose only purpose is to be a stable required-check
context, because the matrix legs report as `test-matrix (os, ver)` and are not fixed
contexts:

```yaml
  test:
    needs: [test-matrix, onnx-export]
    if: always()
    steps:
      - run: |
          [ "${{ needs.test-matrix.result }}" = "success" ] || [ "${{ needs.test-matrix.result }}" = "skipped" ]
          [ "${{ needs.onnx-export.result }}" = "success" ] || [ "${{ needs.onnx-export.result }}" = "skipped" ]
```

`if: always()` means the gate also runs when the workflow run was **cancelled**. Its
upstream results are then `cancelled`, which is neither `success` nor `skipped`, so the
check exits 1 and publishes a **red required status check for a run nobody cares about**.

## Why it bites now

Pushing a stack fires two `pull_request` events per branch — the push itself and the
stack update — and `concurrency.cancel-in-progress` kills the first. That is normal and
harmless in itself. The problem is that the cancelled run leaves a failed `test` / `build`
behind on the same commit as the successful one.

Observed while landing a six-PR stack: two PRs sat at `mergeStateStatus: BLOCKED` with
every real job green, and a third showed two red X's while still merging. Being the more
recent result was **not** enough to clear it — the failed instance had to be removed by
re-running the cancelled workflow (`gh run rerun <id>`).

This will recur on every stack push, and on any other double-trigger.

## The constraint that makes this non-trivial

`always()` is not gratuitous. It is what lets the gate run — and pass — when the matrix
legs are **skipped**, which is exactly what the `changes` path filter does on a doc-only
PR. Replacing it carelessly turns every doc-only PR into a permanently pending required
check, which is worse than the bug.

So the fix has to keep three behaviours simultaneously:

| upstream state | required gate must |
| --- | --- |
| success | report green |
| skipped (doc-only PR) | report green |
| failure | report red |
| **cancelled** | **report nothing at all** |

## Candidate fix

```yaml
    if: ${{ !cancelled() }}
```

`!cancelled()` runs on success, failure and skip, but not when the run was cancelled —
which is the exact distinction wanted. Applies to the `test` gate in `test.yml` and the
`build` gate in `build.yml`.

Worth checking at the same time whether the gate should distinguish `cancelled` from
`failure` in its own upstream test, rather than lumping both into "not success, not
skipped".

## Acceptance

- [ ] A cancelled run publishes no red required check
- [ ] A doc-only PR (matrix legs skipped) still reports the gate green — verify against a
      real doc-only PR, not by reading the expression
- [ ] A genuine matrix failure still reports the gate red
- [ ] Confirmed on a branch that has actually been double-triggered, since that is the
      only shape that reproduces it

## Unrelated note, same area

An open Dependabot PR bumps `astral-sh/setup-uv`, and the SHA-pinning work in this stack
touches the same lines. Whichever lands second needs a trivial rebase.

YeapJieShen:
Closes #161.

## The change

Both workflows' gate jobs move from `if: always()` to `if: ${{ !cancelled() }}`.

`always()` also fires when the **run** was cancelled. The upstream results are then
`cancelled` — neither `success` nor `skipped` — so the gate exits 1 and leaves a red
required check sitting on the same commit as the successful run. Pushing a stack fires two
`pull_request` events per branch and `cancel-in-progress` kills the first, so this recurred
on every stack push.

`always()` was not gratuitous: it is what lets the gate pass when the legs are **skipped**,
which is what the path filter does on a doc-only PR. `!cancelled()` keeps that and drops
only the cancelled case.

An upstream cancelled while the run itself lives (fail-fast killing a sibling) still reports
red — correct, so the shell test is unchanged.

## Acceptance

- [x] A genuine matrix run still reports the gate green — this PR touches `.github/**`,
      which the filter treats as code, so its own matrix is the evidence.
- [ ] A doc-only PR still reports the gate green — **verified by the PR stacked on top of
      this one**, which is `CHANGELOG.md` only and so runs with the legs skipped. That is
      why it is stacked rather than independent: the issue asks for this to be verified
      against a real doc-only PR, not by reading the expression.
- [ ] A cancelled run publishes no red check — observable the next time a stack push
      double-triggers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,7 +91,13 @@ jobs:
   # Stable required-check context: matrix legs are not fixed contexts themselves.
   build:
     needs: [build-matrix, editable-install]
-    if: always()
+    # Not always(): that also fires when the RUN is cancelled, and a cancelled upstream is
+    # neither success nor skipped, so the gate published a red required check for a run
+    # nobody cares about. !cancelled() still runs on success, failure and skip -- the skip
+    # case is a doc-only PR, and is the whole reason this job cannot simply depend on
+    # success. An upstream cancelled while the run itself lives (fail-fast killing a
+    # sibling) still reports red, which is correct.
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
     steps:
       - name: Require all matrix legs to pass or be skipped (doc-only PR)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,7 +155,13 @@ jobs:
   # Stable required-check context: matrix legs are not fixed contexts themselves.
   test:
     needs: [test-matrix, onnx-export]
-    if: always()
+    # Not always(): that also fires when the RUN is cancelled, and a cancelled upstream is
+    # neither success nor skipped, so the gate published a red required check for a run
+    # nobody cares about. !cancelled() still runs on success, failure and skip -- the skip
+    # case is a doc-only PR, and is the whole reason this job cannot simply depend on
+    # success. An upstream cancelled while the run itself lives (fail-fast killing a
+    # sibling) still reports red, which is correct.
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
     steps:
       - name: Require all matrix legs to pass or be skipped (doc-only PR)


YeapJieShen (commented):
Human reviewed, all good.

Issue #157, opened 2026-08-26
Type-check the five modules still exempt from the gate


YeapJieShen:
## Context

Type checking became a required status check, and the exemption list was narrowed
from the glob `sisr.training.*` to five individually named modules. Five remain
exempt:

- `sisr.cli`
- `sisr.training.callbacks`
- `sisr.training.datamodule`
- `sisr.training.gan_module`
- `sisr.training.lightning_module`

Together those are roughly half the package by line count, so the gate currently
covers the other half.

## Why they were exempted

Not neglect. Lightning's hook system takes dynamic `**kwargs` call signatures and a
callback protocol, and jsonargparse resolves subclasses at runtime through
`class_path` dispatch and dotted overrides. Both fight a checker rather than benefit
from one, and annotating around them is a real piece of work rather than a sweep.

## Why it is worth doing anyway

The narrowing already paid for itself once: moving the scoring module out of the
glob type-checked it for the first time and immediately surfaced a genuine
widening, where a library returns a union that the call site had always treated as
the scalar. These five modules have never been checked at all, so the same class of
finding is likely to be sitting in them.

This is also the code-health foundation for adding more architecture families, and
retrofitting it after they land is more expensive than before.

## Suggested approach

Take them one at a time, smallest first, each its own PR — the list is deliberately
structured so that is possible. `datamodule` is probably the cheapest and `cli` the
hardest. Stop and re-scope if a module needs more `type: ignore` comments than it
removes real ambiguity; an exemption honestly argued is better than a file full of
suppressions.

## Acceptance

- [ ] Each module either checks clean or keeps an exemption with a written reason
- [ ] `tests/test_typing.py`'s pinned exemption list updated in the same commit as
      any change to it
- [ ] No net increase in blanket `type: ignore` comments

YeapJieShen:
## Measured, so the suggested order can be corrected

The issue guessed `datamodule` cheapest and `cli` hardest. Half right — `cli` is in fact
the *easiest*. Error counts with each module's exemption lifted in turn, everything else
left exempt:

| module | errors |
|---|---|
| `sisr.training.callbacks` | 39 |
| `sisr.training.gan_module` | 9 |
| `sisr.training.lightning_module` | 9 |
| `sisr.training.datamodule` | 5 |
| `sisr.cli` | 4 |

**`callbacks` is the large one**, and it is also the module an open review candidate wants
split into three. Typing it before that split means typing it twice, so the two are worth
[...]

PR #180, opened 2026-08-29
fix: type-check the datamodule, and refuse a loader built before setup

YeapJieShen:
First of the five modules exempt from the type gate, and the cheapest. Four remain, so
**#157 stays open** — this PR does not close it.

## The finding

Typing the module surfaced exactly the class of defect the narrowing was expected to
surface. The dataset attributes are `Optional` until `setup()` instantiates them, and
`train_dataloader` / `val_dataloader` read them without narrowing.

That is not merely a checker complaint. **`DataLoader` wraps `None` without complaint** —
its sampler only takes `len()` on iteration — so a loader built before `setup()` came back
looking fine and failed later, somewhere else, naming neither the module nor the call that
had been skipped. The test is `DID NOT RAISE` on pre-fix `main`, not a wrong exception type.

## The fix

`_require()` narrows and raises, naming the `setup` stage that fills the dataset. It follows
the shape `predict_dataloader()` already used for its own missing-dataset case, so the
module now answers both questions the same way.

## Evidence

- Test proven RED on pre-fix `main` (`DID NOT RAISE RuntimeError`, both loaders) → GREEN.
- `sisr.training.datamodule` removed from the mypy exemption list; `mypy sisr` clean.
- `ruff` + `ruff format` clean; `tests/training/test_datamodule.py` + `test_distributed.py`
  36 passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,7 +136,6 @@ check_untyped_defs = true
 # Lightning's hook system and jsonargparse's subclass resolution fight a checker.
 module = [
     "sisr.training.callbacks",
-    "sisr.training.datamodule",
     "sisr.training.gan_module",
     "sisr.training.lightning_module",
     "sisr.cli",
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #181, opened 2026-08-29
fix: type-check the CLI

YeapJieShen:
Second of the modules exempt from the type gate. Three remain, so **#157 stays open**.

## A correction worth making in the issue

The exemption list predicted `cli` would be the hardest. Measured, it is the *easiest* —
four errors. Actual counts with each module's exemption lifted in turn:

| module | errors |
|---|---|
| `sisr.training.callbacks` | 39 |
| `sisr.training.gan_module` | 9 |
| `sisr.training.lightning_module` | 9 |
| `sisr.cli` | 4 |
| `sisr.training.datamodule` | 5 (landed below) |

`callbacks` is the large one, and it is also the module an open review candidate wants
split into three — worth sequencing those together rather than typing it twice.

## The changes

Three were missing annotations on hook overrides (`add_arguments_to_parser`,
`before_instantiate_classes`, `__init__`).

The fourth is real but not actionable: mypy cannot rule out a positional `*args` long
enough to reach `LightningCLI`'s own `trainer_class` parameter, so the forwarding call
looks like it passes that argument twice. No caller does this, and narrowing `*args` to
forbid it would mean restating `LightningCLI`'s whole signature here. Suppressed at the one
call site with the reason stated — one ignore against three real annotations, which is the
ratio #157 asks to stay on the right side of.

## Evidence

`mypy sisr` clean with `sisr.cli` removed from the exemption list. `ruff` + `ruff format`
clean. `tests/test_cli.py` 38 passed.

Stacked on #180.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,7 +138,6 @@ module = [
     "sisr.training.callbacks",
     "sisr.training.gan_module",
     "sisr.training.lightning_module",
-    "sisr.cli",
 ]
 ignore_errors = true
 
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -161,7 +161,9 @@ class SRLightningCLI(LightningCLI):
     ``Trainer`` subclass), as long as it also provides ``export``.
     """
 
-    def __init__(self, *args, trainer_class: type[Trainer] = _ExportTrainer, **kwargs):
+    def __init__(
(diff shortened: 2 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #183, opened 2026-08-29
fix: type-check the lightning module, and state the need_sr_rgb contract

YeapJieShen:
Third of the modules exempt from the type gate. Two remain, so **#157 stays open**.

## The interesting half: a contract the annotations could not express

`_forward_lr`, `_forward_sr` and `_step` all return an Optional `sr_rgb` — but whether it
is `None` is decided **entirely by the `need_sr_rgb` argument the caller passed**. So every
consumer already knew it was not `None`, and none of them could say so. Scoring, the public
predict seam and `predict_step` were all reading through an Optional they had themselves
ruled out one call earlier.

Overloading on `Literal[True]` / `Literal[False]` states the real contract, with a third
plain-`bool` variant for the implementations forwarding their own runtime flag. **Three
errors removed with no suppression**, and the signature now says what the docstring already
claimed.

## The rest

- **`nn.Module.__getattr__` is typed `Tensor | Module`**, so `self.model` resolved to a
  union whose `Tensor` arm is not callable. Declaring `model`, `processor`, `criterion` and
  `_compiled` at class level fixes it and is worth having regardless — `_compiled` is
  deliberately **not** a registered submodule (`object.__setattr__` bypasses registration to
  keep one set of weights in `state_dict`), so that declaration is the only type it has ever
  had.
- **`_check_example_input_shape` takes the shape** its caller already branched on, instead
  of re-reading a field that is Optional at the point of use.
- **`trainer.datamodule` is not on `Trainer`'s class surface.** Lightning's data connector
  assigns it dynamically in its `__init__` (to `None`), so it is reached with `getattr` and
  a default. Verified against the installed Lightning rather than assumed — the attribute
  genuinely does not exist statically, and genuinely is always set at runtime.

## Evidence

`mypy sisr` clean with `sisr.training.lightning_module` removed from the exemption list.
`ruff` + `ruff format` clean. `tests/training/test_lightning_module.py` +
`test_gan_module.py` — 147 passed.

Stacked on #181.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,7 +137,6 @@ check_untyped_defs = true
 module = [
     "sisr.training.callbacks",
     "sisr.training.gan_module",
-    "sisr.training.lightning_module",
 ]
 ignore_errors = true
 
--- a/sisr/training/lightning_module.py
+++ b/sisr/training/lightning_module.py
@@ -9,7 +9,7 @@ itself. Optimizer / LR scheduler are wired in from top-level YAML keys by
 """
 
 import dataclasses
-from typing import Any
+from typing import Any, Literal, overload
 
 import lightning
 import torch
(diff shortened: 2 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #184, opened 2026-08-29
fix: type-check the GAN module, and refuse a scheduler it cannot step

YeapJieShen:
Fourth of the five modules exempt from the type gate. **Only `callbacks` remains**, so #157
stays open.

## A real defect, found by typing

`ReduceLROnPlateau.step()` takes a metric. This module steps its schedulers **by hand** in
`on_train_batch_end`, with none to give them, because manual optimization opts out of
Lightning's `monitor` plumbing.

Configuring one did not fail at setup. It failed **part-way into a run**, with a bare
`TypeError` about a missing positional argument — after however many hours of training had
already happened.

`configure_optimizers` now refuses it at startup and names what to use instead (the paper's
own `1e-4 -> 1e-5` drop is time-based anyway). `_time_schedulers` re-checks rather than
trusting a guarantee it does not own, since a subclass can replace `configure_optimizers`.

## The bulk: one missing declaration

Six of the nine errors were the same thing. `SRGANLightning.__init__` already refuses a
`training_config` that is not an `SRGANTrainingConfig` — but the attribute kept the base
class's type, so every adversarial-only field it reads (`init_from`, `adversarial_weight`,
`d_steps_per_g_step`) looked undefined. Declaring the narrowed type records a guarantee the
class already enforces.

## Also

`_paired_optimizers` turns "there are exactly two optimizers, generator first" from an
invariant `training_step` assumed into one it checks — `optimizers()` spans the
single-optimizer case in its return type, and a subclass returning a different shape would
otherwise fail as a confusing unpack.

**One ignore remains, deliberately**: `training_step` returns `None`. Lightning's own
contract permits that under manual optimization; only this project's base class narrows the
return to `Tensor`.

## Evidence

- Four tests proven RED on pre-fix code, GREEN after.
- `mypy sisr` clean with `sisr.training.gan_module` removed from the exemption list.
- `ruff` + `ruff format` clean; `tests/training/test_gan_module.py` 43 passed.

Stacked on #183.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,7 +136,6 @@ check_untyped_defs = true
 # Lightning's hook system and jsonargparse's subclass resolution fight a checker.
 module = [
     "sisr.training.callbacks",
-    "sisr.training.gan_module",
 ]
 ignore_errors = true
 
--- a/sisr/training/gan_module.py
+++ b/sisr/training/gan_module.py
@@ -18,6 +18,7 @@ from typing import Any
 
 import torch
 from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+from lightning.pytorch.core.optimizer import LightningOptimizer
 from lightning.pytorch.utilities.types import OptimizerLRScheduler
 from lightning_utilities.core.rank_zero import rank_zero_info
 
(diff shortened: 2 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #189, opened 2026-08-29
docs: correct every stale .pt reference the artifact arc left behind

YeapJieShen:
The arc removed `.pt` entirely. **Fifteen references in tracked files still named it.**

## One of them reaches users

`SRGANTrainingConfig.init_from`'s docstring is what jsonargparse renders into `--help`. It
said *"Path to a bare-weights `.pt`"* — so the CLI was telling a reader to pass a file format
the code now refuses to load.

## One contradicted itself three lines apart

```yaml
# Bare weights (.pt), never the sibling .ckpt. Mismatches are refused.
example_input_shape: [3, 24, 24]
init_from: "…/SRResNet_x4_RGB_16B64F_s1000000.safetensors"
```

## The rest

Docstrings in `callbacks.py` (4), `gan_module.py`, `metadata.py` (2), and test docstrings
that described the artifact by an extension it no longer has.

Also renamed the metadata tests' temp files from `meta.pt` to `meta.ckpt`. Those round-trip
through `torch.save`/`torch.load`, whose sink is the **resumable checkpoint**, not the bare
artifact — so `.pt` was doubly wrong there and `.ckpt` is what it actually is.

## Deliberately kept

`docs/configuration.md` and `metadata.py`'s `FORMAT_VERSION` comment both state that the
`.pt` form no longer exists, and the CHANGELOG records its removal. Those are *about* the
format rather than *in* it.

## Evidence

`ruff` + `ruff format` + `mypy sisr` clean. `tests/test_cli.py` + `tests/training/test_metadata.py`
63 passed. Re-scanned tracked files afterwards: the only `.pt` left is in the three places
named above.

### DIFF
--- a/sisr/models/srgan/config.py
+++ b/sisr/models/srgan/config.py
@@ -14,7 +14,7 @@ class SRGANTrainingConfig(SRResNetTrainingConfig):
     """SRGAN training defaults. The generator is SRResNet, so ``scale=4`` is inherited.
 
     Args:
-        init_from: Path to a bare-weights ``.pt`` whose generator weights
+        init_from: Path to a bare-weights ``.safetensors`` whose generator weights
             initialise this run's generator. The paper scopes the MSE-init
             trick to "when training the actual GAN", so a paper-faithful SRGAN
             run starts from an MSE-trained SRResNet. Optional: unset trains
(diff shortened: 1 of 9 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #190, opened 2026-08-29
fix: an artifact save creates the directory it is given

YeapJieShen:
`SRWeightsCheckpoint` could not write into a directory nothing else had made.

## Why it never surfaced

Lightning creates a checkpoint's directory inside its own IO plugin
(`fabric/plugins/io/torch_io.py`, `fs.makedirs`). `SRWeightsCheckpoint._save_checkpoint`
**deliberately bypasses that path** to write a different payload — bare safetensors rather
than a full checkpoint — so nothing was creating the directory.

It worked only because every shipped template pairs it with an `SRCheckpoint`, which uses
Lightning's own save path and makes the directory first. **The dependency was invisible and
unstated.**

A weights-only configuration — or one whose weights callback fires first — dies on its very
first save with a bare `SafetensorError` about a temp file, naming neither the callback nor
the missing directory.

## How it was found

A smoke test that added a third checkpoint callback firing on a step boundary
(`every_n_train_steps`) rather than on validation, so it ran before any `SRCheckpoint` had
created the directory.

## Where the fix goes

`artifacts.save`, not the callback. That function owns writing the artifact, so it owns
making the destination writable; the caller still owns the filename. Putting it in the
callback would leave the same trap for the next caller — `sisr export` writes through the
same function.

## Evidence

- Test proven RED on pre-fix `main` (`SafetensorError`) → GREEN.
- `ruff` + `ruff format` + `mypy sisr` clean.
- `tests/test_artifacts.py` + `tests/training/test_callbacks.py` + `tests/test_typing.py`:
  128 passed. (`test_typing.py` included deliberately — it is the guard I missed on an
  earlier batch.)

### DIFF
--- a/sisr/artifacts.py
+++ b/sisr/artifacts.py
@@ -83,10 +83,16 @@ def save(path: str | PathLike, tensors: dict[str, torch.Tensor], meta: dict[str,
     """Write one component's weights plus its provenance.
 
     Args:
-        path: Destination. The caller owns the filename.
+        path: Destination. The caller owns the filename; this owns the directory.
         tensors: A component's ``state_dict()``.
         meta: Provenance to carry in the header.
     """
+    # Lightning creates a checkpoint's directory inside its own IO plugin, and
+    # SRWeightsCheckpoint._save_checkpoint deliberately bypasses that path to write a
+    # different payload -- so nothing was creating this one. It worked only while a sibling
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #193, opened 2026-08-30
fix: refuse two checkpoint callbacks that would write the same filename

### LINKED ISSUE #191: Two checkpoint callbacks with different monitors write the same filename
## The defect

Two checkpoint callbacks of the same class monitoring **different** metrics now write the
**same filename**, and Lightning disambiguates with a `-v1` suffix assigned by save order.

The shipped templates configure exactly this: `SRCheckpoint` on `psnr/val/RGB` and another on
`ssim/val/RGB`, plus the same pair of `SRWeightsCheckpoint`. Observed in a real fit:

```
SRResNet_x4_RGB_16B64F_s19.ckpt
SRResNet_x4_RGB_16B64F_s19-v1.ckpt
SRResNet_x4_RGB_16B64F_s19.safetensors
SRResNet_x4_RGB_16B64F_s19-v1.safetensors
```

**Nothing in the filename says which is the PSNR-best and which is the SSIM-best.**

## Why it appeared

The artifact-filename change gave every artifact a descriptive stem —
`{Arch}_x{scale}_{colorspace}_{variant}_s{step}` — and dropped the metric from the name. Before
that, filenames carried the monitored metric and its value
(`sr-weights-1000000-psnr_val_RGB=28.8635`), so two monitors could never collide.

The class docstring anticipates *cross-class* collisions and solves them with distinct default
prefixes: *"each callback only ever names/deletes files matching its own prefix"*. That reasoning
does not cover two instances of the **same** class.

## Why it matters

- **The `-v1` assignment is save order, not identity.** Which callback wins the bare name is not
  guaranteed stable across runs or Lightning versions.
- **The rolling-window and top-k bookkeeping match on prefix.** Two callbacks sharing a prefix in
  one directory is the shape that caused an earlier eviction bug, where a directory holding two
  numbering schemes let a run delete its own newest saves.
- A published artifact whose filename cannot say what it is defeats the point of naming it
  descriptively in the first place.

## Workaround in use

The 2026-08-30 reference retrain sets an explicit `filename_prefix` on the SSIM-monitored
callbacks and leaves the PSNR ones on the derived stem — so the artifact `init_from` expects is
still produced under exactly its expected name, and the SSIM files are distinguishable. That is a
per-config workaround, not a fix.

## Candidate fixes

1. **Include the monitor in the derived filename** when `monitor` is set — closest to the old
   behaviour, and self-describing.
2. **Refuse two callbacks that would derive the same filename in the same `dirpath`**, at setup,
   naming both. Cheap, and turns a silent ambiguity into a startup error.

(1) and (2) are not exclusive, and (2) is worth having regardless — it also covers a user who
configures two callbacks with the same explicit prefix.

## Acceptance

- [ ] Two same-class callbacks with different monitors produce distinguishable filenames, or are
      refused at setup with a message naming both
- [ ] A test asserts it, proven RED against current `main`
- [ ] The reference config no longer needs a `filename_prefix` workaround

YeapJieShen:
Closes #191.

## What changed

Two checkpoint callbacks that would write the same filename into the same directory are now
refused at `setup()`, with a message naming both callbacks, the filename they share, and how to
resolve it. Both shipped templates now give their SSIM-monitored callbacks a distinct
`filename_prefix`.

## Why

The templates configured exactly the colliding shape: `SRCheckpoint` on `psnr/val/RGB` beside one
on `ssim/val/RGB`, plus the same pair of `SRWeightsCheckpoint`. Both derive their name from the
model's provenance, which says nothing about the selection criterion, so both land on one stem and
Lightning keeps one and appends `-v1` to the other — **assigned by save order**. The published
artifacts then cannot say which is best on which metric, and which file wins the bare name is
stable across neither runs nor Lightning versions.

Distinct default prefixes already kept the two *classes* apart, and their docstrings said so. That
reasoning never covered two instances of one class.

### Why refuse rather than disambiguate automatically

Any automatic rule has to decide which sibling keeps the bare name, and every answer is either
order-dependent — which is the defect itself — or silently renames published artifacts when a
monitor changes.

Putting the metric back into the derived name (issue #191's option 1) would be better still, and
this change does not foreclose it. It is deliberately not done here: it changes a filename that a
just-merged template's `init_from` names and that a completed reference run has already written on
disk, so it wants to be one deliberate rename, coordinated with those, rather than a side effect.
**Issue #191 stays open for that**; its third acceptance box — "the reference config no longer
needs a `filename_prefix` workaround" — is not ticked by this PR, and the prefix is now required
rather than merely advisable.

## Evidence

**Bug fix** — proven RED against the unfixed code, GREEN after, at implementation time:

```
$ pytest tests/training/test_callbacks.py -k "<the seven new tests>"
3 failed, 4 passed          # against unmodified main
115 passed                  # after the fix (108 pre-existing + 7 new)
```

The three that failed did so with `DID NOT RAISE MisconfigurationException`:

- `test_two_checkpoints_with_different_monitors_are_refused_at_setup`
- `test_the_refusal_does_not_depend_on_which_sibling_lightning_sets_up_first`
- `test_the_refusal_names_the_filename_the_pair_would_share`

The other four were written to pass *both* ways on purpose — they pin what must stay legal, so an
over-broad check is caught:

- `test_a_distinct_prefix_on_one_of_the_pair_is_enough`
- `test_the_two_classes_may_share_a_prefix_because_their_extensions_differ` — `.ckpt` beside
  `.safetensors` under one prefix is the intended, shipped pair, so the comparison is on the whole
  filename and not the prefix
- `test_weights_checkpoints_on_different_components_do_not_collide` — the SRGAN template's
  generator/discriminator pair, same class and same null monitor, must stay constructable
- `test_separate_dirpaths_are_not_a_collision`

Full suite: **404 passed** (`tests/training/`, `test_typing.py`, `test_artifacts.py`) plus the
remainder green. `ruff check` and `ruff format --check` clean.

## Notes

- The check is symmetric, so it does not matter which sibling Lightning sets up first.
- It compares the *raw* `dirpath` rather than the resolved one, because `setup()` resolves
  `self.dirpath` against the trainer — when one callback probes another, one side may be resolved
  and the other not.
- A sibling whose own configuration is broken is skipped rather than probed, so its own `setup()`
  reports that precisely instead of being masked by a collision error raised from elsewhere.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -770,6 +770,10 @@ class _SRCheckpointBase(_RollingSaveMixin, ModelCheckpoint):
         )
         self._keep_last = keep_last
         self._explicit_prefix = filename_prefix
+        # The *raw* dirpath, because setup() resolves self.dirpath against the
+        # trainer -- so when one callback probes another, one side may be resolved
+        # and the other not, and comparing those is comparing different things.
+        self._explicit_dirpath = None if dirpath is None else str(Path(dirpath).resolve())
         # A usable name before setup() can derive the real one -- direct
         # construction in a test never reaches setup.
         self._apply_naming(filename_prefix or self.DEFAULT_PREFIX)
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #194, opened 2026-08-30
refactor: type-check sisr.training.callbacks, the last exempt module

YeapJieShen:
The exemption list is now empty, and `EXPECTED_IGNORED` says so. The stated
reason -- Lightning's hook system and jsonargparse's subclass resolution fight a
checker -- was real but narrower than a blanket `ignore_errors`. It came down to
three seams:

Lightning types every hook as taking a bare `LightningModule`, while everything
these callbacks read from it (`eval_config`, `scorer`, `predict_rgb`, the
provenance builders) belongs to `SRLightning`. Overriding a hook with a narrower
parameter is unsound and rejected, so the narrowing happens once, at a named
seam, rather than being asserted everywhere or ignored.

`_RollingSaveMixin` reads `monitor`, `dirpath`, `FILE_EXTENSION` and
`_remove_checkpoint` off the `ModelCheckpoint` it is combined with, and chains
`on_train_start` through it -- none of which a standalone mixin can declare. It
now names that base for the checker only, so the runtime MRO stays exactly what
its docstring describes.

`dataset_names` was `list[str] | None` where None and `[]` already reached the
same branch, so the Optional bought nothing but a re-narrow at every read. The
two remaining Optionals are Lightning's own stage-agnostic dataloader
declarations, asserted where the hook that reads them cannot fire without them.

No behaviour changes: `mypy sisr/` passes over all 48 files with no exemptions,
and the callbacks suite is unchanged and green.

Stacked on #193 — merge that first.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,13 +133,6 @@ disallow_untyped_defs = true
 check_untyped_defs = true
 
 [[tool.mypy.overrides]]
-# Lightning's hook system and jsonargparse's subclass resolution fight a checker.
-module = [
-    "sisr.training.callbacks",
-]
-ignore_errors = true
-
-[[tool.mypy.overrides]]
 # Neither ships inline types or a py.typed marker.
 module = ["torchvision.*", "tqdm.*"]
 ignore_missing_imports = true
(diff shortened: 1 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #196, opened 2026-08-30
feat: let a config select torch.compile's mode, not just its backend

YeapJieShen:
Makes `torch.compile`'s `mode` selectable from a config. It was not, and that quietly
invalidated a class of benchmark figure.

## What changed

- `SRTrainingConfig.compile_mode`, passed to `torch.compile` alongside `compile_backend`.
- Defaults to `None`, which is what `torch.compile` already used — **no existing compiled run
  changes**.
- Requires `compile_backend='inductor'`, refused at config construction otherwise.
- The `compile_backend` docstring's measurements were stale and are corrected.

## Why

The call site passed `backend` and nothing else, so every configured run got inductor's *default*
mode. Every `reduce-overhead` / `max-autotune` figure in the benchmark record was produced by
patching the mode into a benchmark process — describing something no YAML could select. On this
hardware the gap between default mode and `reduce-overhead` is around 15% at fp32, so it is not a
rounding difference.

## Why it requires inductor

`mode` is an inductor setting. With no backend nothing is compiled and the mode is dead config.
With any *other* backend, torch forwards `mode` as a keyword to a compiler function that does not
accept one — which fails on the **first compiled call**, not at startup. This project's compile
plumbing (the `on_fit_start` warm-up, the eager `InvalidBackend` check) exists precisely to turn
that class of late failure into an early one, so the config refuses the combination instead.

Verified against torch 2.13.0: inductor validates the mode name eagerly at `torch.compile()`,
while non-inductor backends accept any string and defer.

## Evidence

Six tests, all proven RED before the implementation (`TypeError: unexpected keyword argument
'compile_mode'`). Full suite: **854 passed, 5 skipped**.

## Note

This does not recommend compiling. #121's finding stands — no backend is bit-identical to eager on
SRResNet. The value here is that the benchmark record becomes reproducible from a config.

### DIFF
--- a/sisr/training/config.py
+++ b/sisr/training/config.py
@@ -121,12 +121,31 @@ class SRTrainingConfig:
             ``SRLightning.on_fit_start`` turns into an immediate failure via
             a warm-up forward instead of an arbitrary mid-run crash.
             Measured on one RTX 5060 Laptop (SRResNet, batch 16, 24x24 LR):
-            ``'cudagraphs'`` gave +4% steps/s over eager — it only captures
-            the model forward, so backward and the optimizer step stay
-            eager and most kernel launches in a training step are never
-            captured. ``'inductor'`` fails outright there (no upstream
-            Windows Triton wheel). Defaults to ``None`` so this mostly-
-            unproven-on-this-project path never ships on by default.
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

Issue #199, opened 2026-08-31
Settle what the papers leave unspecified, from primary sources


YeapJieShen:
## Goal

The README now states four places where the source papers stop specifying. Each is currently
labelled honestly but rests on an inference or on nothing. **Resolve as many as a primary source
can settle, and sharpen the label on whatever survives.**

The standing principle from the section itself holds and is not up for revision here: resolving a
row means **finding the authoritative source**, never picking a plausible value and asserting it.
A row that no source can settle stays labelled as unsettled — that is a success outcome, not a
failure.

### 1 + 2. SRCNN batch size and training budget — one artifact settles both

Currently: batch size `64` rests on **nothing**; the budget (`max_steps: 10000000`) rests on
reading the paper's *"the same number of backpropagations (i.e. 8x10^8)"* as **per-sample**
rather than per-mini-batch, and an earlier reading of ours recorded *"up to 8x10^8"* — a ceiling,
not a target. The two readings are unreconciled.

The authors released Caffe training files. A solver prototxt carries `max_iter`, and the train
prototxt carries the data layer's `batch_size`. **Those two numbers together settle both rows and
also adjudicate the per-sample vs per-mini-batch reading**, because `max_iter x batch_size` either
lands on 8x10^8 or it does not.

Task: locate the authors' released Caffe files, record where they were obtained, quote the two
fields verbatim, and reconcile against the paper sentence. If they disagree with what ships,
**do not silently change the config** — report the discrepancy and let the change be a separate,
argued decision, since it would invalidate the existing SRCNN numbers.

### 3. SRCNN evaluation border

Currently: `crop_border: 3` on top of the 6 px per side `padding: valid` already trims. The paper
states no test-time boundary convention.

The authors also released evaluation/demo code. If it shaves a fixed border, that is the
convention the paper's own numbers were produced under, and it is checkable. Failing that, record
what the dominant convention in the SISR literature is **and say it is a convention rather than
the paper's**.

### 4. SRResNet PReLU parameterisation

Currently: `torch.nn.PReLU()`, one shared slope. The paper specifies neither shared nor
per-channel and there is no official implementation.

Least likely to be settled, and the README already says community reimplementations are not a
substitute. Two things can still sharpen it: what the PReLU paper itself uses as its default, and
whether the SRResNet parameter count implied anywhere in the paper distinguishes the two (a
per-channel slope adds one parameter per channel per activation, which is a countable difference).
[...]

PR #206, opened 2026-09-01
fix: use the SRCNN authors' own batch size and training budget

### LINKED ISSUE #202: SRCNN batch size and training budget: settle both from the authors' Caffe files
Sub-issue of #199. Rows 1 and 2 of the README's "Where the papers stop specifying" section.
They are one task because a single artifact settles both.

## What ships, and what it rests on

| row | ships | rests on |
|---|---|---|
| SRCNN batch size | `batch_size: 64` | **nothing checkable** |
| SRCNN training budget | `max_steps: 10000000` | an inference, plus two unreconciled readings |

The budget reads the paper's *"the same number of backpropagations (i.e. 8x10^8)"* as
**per-sample** rather than per-mini-batch. An earlier reading of ours recorded *"up to
8x10^8"* — a ceiling, not a target. Those two readings have never been reconciled.

## Why one artifact settles both

The authors released Caffe training files. A solver prototxt carries `max_iter`; the train
prototxt carries the data layer's `batch_size`. **`max_iter` x `batch_size` either lands on
8x10^8 or it does not**, which adjudicates the per-sample vs per-mini-batch reading at the
same time as it fixes the two values.

Splitting these into separate issues would invite closing one on half the evidence.

## Done when

One of:

- **Settled** — both fields quoted verbatim, the source and where it was obtained recorded,
  and the paper sentence reconciled against them.
- **Narrowed** — the space of defensible values shrank, and the README says how and why.
- **Unsettleable** — the search is written down so nobody repeats it, and both rows stay
  labelled as resting on an inference.

An explicit "we looked at X and it does not say" is a success outcome, not a failure.

## If the source disagrees with what ships

Change the config to the authors' values, then **file the re-run separately** — it invalidates
every published SRCNN figure, and the superseded numbers move to the benchmark archive stamped
with the config that produced them rather than being retracted. Do not fold the re-run into
this issue.

YeapJieShen:
Closes #202. Bottom of the #199 stack.

## What settles it

The SRCNN authors released a Caffe training package. It was named in #199 as the artifact
that would settle rows 1 and 2 of the README's "Where the papers stop specifying" table, and
it does — both, from one file each.

- **`SRCNN_train.zip`**, from the CUHK project page linked by both the ECCV 2014 and TPAMI
  2015 papers
- retrieved **2026-09-01**, SHA-256
  `001146419f7acfb12a3e7929c8acd5de88a08d687d6881085f81321ad6982b1a`

## The two rows

| field | authors | shipped before | shipped now |
|---|---|---|---|
| `hdf5_data_param.batch_size` (TRAIN phase) | `128` | `64` | **`128`** |
| `max_iter` | `15000000` | `max_steps: 10000000` | **`max_steps: 15000000`** |

## The paper and the authors' code disagree, and this preserves that

The paper says *"the same number of backpropagations (i.e. 8x10^8)"*. The solver says
`max_iter: 15000000`. At batch 128 that is **1.92x10^9** — a factor of **2.4** apart.

**Settled by the file:** 8x10^8 cannot mean *iterations*, since it exceeds the authors' own cap
by two orders of magnitude. Per-sample is the only consistent reading, which this project had
already inferred and no longer has to.

**Deliberately left open:** which number is the training budget. This ships the authors'
`15000000`, because that is what a primary source states.

An earlier revision of this branch shipped `6250000` — the value that makes the paper's 8x10^8
come out exact at batch 128. That was **cherry-picking one artifact**: it took `batch_size` from
the solver as authoritative and then overrode `max_iter` from the same file, on the strength of
our own inference that `max_iter` paired with `snapshot: 500` is a ceiling rather than a target.
That inference is plausible, and **which snapshot produced the published numbers is recorded
nowhere**, so it stays an inference. Following the file and stating the disagreement is the
honest position; resolving it silently is not.

Practically this is a long run. Stop on a checkpoint rather than lower the config to a number
no source states.

## Five fields that already matched

Worth stating, because it is evidence the port was faithful where it could be checked:
`base_lr: 0.0001`, `momentum: 0.9`, `weight_decay: 0`, the per-layer `lr_mult` structure
(`1`, `1`, `0.1` -> our `layer_lrs: [1e-4, 1e-4, 1e-5]`), and `weight_filler` std `0.001`.

## Blast radius

**No published number changes** — SRCNN has no publishable row, as the README already states.
`val_check_interval` scales with `max_steps` to keep 20 validation cycles.

One divergence *was* found while comparing, and is filed separately rather than fixed here
because it needs a schema change rather than a value change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,15 +340,25 @@ both rows at once:
 | per-layer `lr_mult` (weights) | `1`, `1`, `0.1` | already matched — `layer_lrs: [1e-4, 1e-4, 1e-5]` |
 | `weight_filler` std | `0.001` | already matched |
 
-`max_iter` also settles how to read the paper's *"the same number of backpropagations
-(i.e. 8×10⁸)"*. It **cannot** mean iterations: 8×10⁸ exceeds the authors' own cap of 1.5×10⁷
-by two orders of magnitude. Per-sample is the only consistent reading — which is what this
-project had already inferred, and it is no longer an inference. Nor is `max_iter` the paper's
-budget: paired with `snapshot: 500` it is a run-long ceiling to snapshot from, which is what
-an earlier reading of *"up to 8×10⁸"* was picking up on. Both readings reconcile.
+`max_iter` settles one thing and opens another.
 
-`max_steps` is therefore **6,250,000** — 6,250,000 × 128 = 8×10⁸ backpropagations exactly,
-where before it was an approximation at ≈0.8× the paper's figure.
+**Settled:** the paper's *"the same number of backpropagations (i.e. 8×10⁸)"* **cannot** mean
+iterations — 8×10⁸ exceeds the authors' own cap of 1.5×10⁷ by two orders of magnitude. Per-sample
+is the only consistent reading, which this project had already inferred and no longer has to.
+
+**Open, and deliberately left open:** at batch 128 the authors' `max_iter` is
+15,000,000 × 128 = **1.92×10⁹** backpropagations, against the paper's stated **8×10⁸** — which
+would be 6,250,000 iterations. **The released code and the paper disagree, by 2.4×.**
+
+This project ships the authors' value, `max_steps: 15000000`, because that is the number a
+primary source states. Choosing 6,250,000 instead would make the paper's sentence come out exact
+at the cost of shipping a value no source states, resolved by our inference that `max_iter` is a
+ceiling rather than a target. Their `snapshot: 500` makes that inference plausible — but **which
+snapshot produced the published numbers is recorded nowhere**, so it stays an inference, and the
+honest position is to follow the file and say the two disagree.
+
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

YeapJieShen:
Corrected on your challenge: this now ships the authors' `max_iter: 15000000`, not the `6250000` that made the paper's 8x10^8 come out exact.

Taking `batch_size` from the solver as authoritative and then overriding `max_iter` from the same file was cherry-picking one artifact, and it replaced an old inference with a new one while presenting the result as settled. The disagreement between the released code (1.92e9 backpropagations at batch 128) and the paper (8e8) is a **finding to preserve**, not something to resolve by picking whichever number reads cleanly.

`val_check_interval` scales with it to keep 20 validation cycles. #208 and #216 are rebased on the correction; #216's `docs/reproduction.md` carries it.

Issue #207, opened 2026-09-01
SRCNN inference padding and evaluation border both differ from the authors' demo code


YeapJieShen:
Found while settling #203 from the SRCNN authors' released demo code
(`SRCNN_v1.zip`, retrieved 2026-09-01, SHA-256
`bfa68ca613c1326a59e0c34353205a254ab2b67e34df7f04e28eef567980af30`).

**This needs a decision before anything is changed.** Two divergences are entangled, and
fixing either alone moves us further from the authors, not closer.

## What the authors do

`demo_SR.m`:

```matlab
im_h   = SRCNN(model, im_b);
im_h   = shave(uint8(im_h * 255), [up_scale, up_scale]);
im_gnd = shave(uint8(im_gnd * 255), [up_scale, up_scale]);
```

and `SRCNN.m` convolves with:

```matlab
imfilter(im_b, weights_conv1(:,:,i), 'same', 'replicate')
```

So at **test time** the network is run with **SAME padding and replicated borders**, producing
a full-size output, and then **`scale` pixels per side** are shaved before PSNR.

Note this differs from their own **training** setup, which is `pad: 0` on 33x33 patches
against 21x21 labels — valid convolution. The authors train valid and test same+replicate.

## What we do

- inference uses `padding: valid`, so the output is already 12 px smaller per dimension
  (6 per side, from kernels 9/1/5)
- then `crop_border: 3` on top

## Why they are entangled

At scale 3, on an HxW ground truth:

| | region scored |
|---|---|
| authors | rows `4 .. H-3` (shave 3 from a full-size output) |
| ours | rows `10 .. H-9` (6 lost to valid conv, then 3 cropped) |

Ours is strictly **inside** theirs, and the two are not the same measurement. Separately,
**`crop_border: 3` is a constant where the authors' border is `scale`** — so our value
coincides with theirs at x3 and disagrees at x2 (theirs 2) and x4 (theirs 4). SRCNN's headline
results are at x3, which is very likely why 3 looked right.

Changing `crop_border` to `scale` on its own would **not** converge on the authors' region,
because the valid-conv loss is still there underneath.

## The decision

1. **Adopt the authors' test-time path** — same + replicate padding at inference, border
   `scale`. Reproduces their measurement exactly. Costs a train/test asymmetry in our own
   code, and means scoring pixels predicted from invented (replicated) border content.
2. **Keep valid + document the divergence.** Arguably more honest — we never score a pixel
[...]

Issue #204, opened 2026-09-01
SRResNet PReLU: shared or per-channel slope


YeapJieShen:
Sub-issue of #199. Row 4 of the README's "Where the papers stop specifying" section, and the
least likely of the four to be settled.

## What ships, and what it rests on

`torch.nn.PReLU()` — one shared slope across all channels. The paper specifies **neither**
shared nor per-channel, and there is no official reference implementation. Community
reimplementations are not a substitute and the README already says so.

## Two routes that can still sharpen it

1. **What the PReLU paper itself uses as its default.** Weak evidence, but it is the
   convention a reader of that era would have assumed.
2. **The parameter count.** A per-channel slope adds one parameter per channel per
   activation, so if the paper implies a parameter count anywhere, the two variants are a
   **countable** difference. This is the interesting one — it is the only route here that
   could turn an unsettleable row into a settled one from the paper alone, with no external
   artifact required.

Start with 2. It is cheap, self-contained, and either resolves the row or rules itself out.

## Done when

One of: **settled** (parameter count or a primary source distinguishes them), **narrowed**
(the space of defensible readings shrank, and the README says how), or **unsettleable** (both
routes written up as exhausted, so nobody repeats them, and the row stays labelled).

YeapJieShen:
Narrowed by PR #208, and the deciding route is now closed with a measurement rather than left un-run. Shared vs per-channel differ by **1,197 parameters, 0.0773% of the model** (1,549,462 vs 1,550,659; 19 PReLU sites, all 64 channels) -- below the precision any paper reports a parameter count at, so no count Ledig et al. could have published would distinguish them. The PReLU paper presents both variants as viable and comparable, so that route does not decide it either. The row stays labelled; both searches are recorded.

YeapJieShen:
Closing as **unsettleable, with both routes exhausted and recorded** — one of the three done-states
this issue defined, and a success outcome rather than a failure.

**Route 2 (the parameter count) is closed with a measurement, not an assumption.** Measured on
this project's SRResNet — 19 PReLU sites, all 64 channels:

| | PReLU params | model total |
|---|---:|---:|
| one shared slope (ships) | 19 | 1,549,462 |
| per-channel | 1,216 | 1,550,659 |

**0.0773% of the model.** No parameter count a paper reports could distinguish them, so this
route could never have settled the row — it is ruled out, not merely un-run.

[...]

PR #208, opened 2026-09-01
docs: settle the SRCNN evaluation border and close the PReLU parameter-count route

### LINKED ISSUE #203: SRCNN evaluation border: find the authors' test-time convention
Sub-issue of #199. Row 3 of the README's "Where the papers stop specifying" section.

## What ships, and what it rests on

`crop_border: 3` at evaluation, on top of the 6 px per side that `padding: valid` already
trims. **The paper states no test-time boundary convention at all**, so this is our choice
presented as a number.

## The tractable route

The authors released evaluation/demo code alongside the training files. If it shaves a fixed
border, that is the convention the paper's own published numbers were produced under, and it
is directly checkable.

## Done when

One of:

- **Settled** — the authors' evaluation code quoted, and the README row rewritten from "rests
  on" to "specified by".
- **Narrowed** — record what the dominant convention in the SISR literature is, **and say
  explicitly that it is a convention rather than the paper's**. That distinction is the whole
  point of the section.
- **Unsettleable** — the search is written down and the row stays labelled.

## Why it matters

An evaluation border changes the reported PSNR/SSIM without changing the model. Two
implementations differing here look like a reproduction failure and are not one — which is
exactly the class of difference this section exists to make visible.

YeapJieShen:
Closes #203 and #204. Stacked on #206.

## #203 — the border is specified, and we differ

The authors' demo package settles it (`SRCNN_v1.zip`, retrieved 2026-09-01, SHA-256
`bfa68ca613c1326a59e0c34353205a254ab2b67e34df7f04e28eef567980af30`):

```matlab
im_h   = shave(uint8(im_h * 255), [up_scale, up_scale]);
im_gnd = shave(uint8(im_gnd * 255), [up_scale, up_scale]);
```

The border is **`scale` pixels per side**, not a constant. And `SRCNN.m` runs the network
with `imfilter(..., 'same', 'replicate')` — SAME padding at test time — although training is
`pad: 0`.

We use `padding: valid` at inference plus `crop_border: 3`. At x3 that scores rows
`10 .. H-9` against their `4 .. H-3`: strictly inside, not the same measurement. Our constant
also agrees with `scale` only at x3, which is very likely why 3 looked right — SRCNN's
headline results are at x3.

**Nothing is changed here.** The two differences are entangled: setting `crop_border: scale`
alone moves us *further* from the authors, because the 6 px per side lost to valid convolution
is still underneath. The trade — reproduce their measurement, or keep a path that never scores
a pixel predicted from invented border content — is #207.

## #204 — narrowed, and the deciding route is closed with a measurement

The parameter count was the only route that could settle this from the paper alone. Measured
on this project's SRResNet — 19 PReLU sites, all 64 channels:

| | PReLU params | model total |
|---|---:|---:|
| shared (ships) | 19 | 1,549,462 |
| per-channel | 1,216 | 1,550,659 |

**0.0773% of the model.** No parameter count a paper reports could distinguish them, so the
route is closed rather than merely un-run. The weaker route fails too: the PReLU paper
presents channel-wise and channel-shared as both viable and comparable, so there is no single
"what the authors would have assumed" either.

The row stays labelled, and both searches are written down so nobody repeats them.

## Net effect

The papers-silent table is **down from four rows to one**, and the section now says so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,21 +307,22 @@ each other, so none are published. A comparable SRCNN row needs a fresh run.
 ### Where the papers stop specifying
 
 A claim that these numbers can be independently verified is only honest if the places the
-source papers **stop specifying** are stated as plainly as the places they do. Each of the
-following is a choice this implementation had to make and **cannot be scored as correct**,
-because there is nothing to be faithful to. None is a bug, and none should be "fixed" by
-picking a value and asserting it in code — that would dress a guess as a checked fact.
+source papers **stop specifying** are stated as plainly as the places they do. Such a choice
+**cannot be scored as correct**, because there is nothing to be faithful to. It is not a bug,
+and it should not be "fixed" by picking a value and asserting it in code — that would dress a
+guess as a checked fact.
+
+**This table held four rows until 2026-09-01. One remains.** What happened to the other three
+is recorded below, because a search that closed a row is worth as much as the row was.
 
 | choice | what ships | what it rests on |
 |---|---|---|
-| SRCNN evaluation border | `crop_border: 3`, on top of the 6 px per side that `padding: valid` already trims | The paper states no test-time boundary convention. Neither value is correct or incorrect, only stated. |
-| SRResNet PReLU parameterisation | `torch.nn.PReLU()` — one shared slope | The paper specifies neither shared nor per-channel, and no official reference implementation exists to check against. |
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #211, opened 2026-09-01
refactor: compact prose in sisr/training, losing no rule

YeapJieShen:
First PR of the prose-compaction stack. **Do not merge without reading the ledger** — that is
what it is for.

`sisr/training` carries **46% of the package's prose** (2,348 of 5,120 lines), so it goes first.

| | before | after |
|---|---:|---:|
| `sisr/training` | 1.80 : 1 | **1.66 : 1** |
| whole package | 1.73 : 1 | **1.67 : 1** |

## Constraint ledger

25 blocks across 7 files. Every removed block is one of:

**compressed (23).** Measurement essays to their conclusion; run-on explanations to the
constraint they carry.

- `config.py` — `compile_backend`'s four-backend benchmark narrative becomes the three facts a
  caller acts on: only inductor pays and by a mid-teens percent, nothing is bit-identical to
  eager, stay at `None`. `compile_mode` keeps its whole refusal rule and the
  "measure per configuration" warning.
- `callbacks.py` — `BenchmarkImageLogger` 62 -> 49 lines with **every** tag name, both
  dataloader-index conventions, the `crop_border` anti-drift rule and the 0.5 GB buffering
  lesson intact. `SRWeightsCheckpoint` 55 -> 49. Nine private docstrings lose trivial
  `Args:`/`Returns:` blocks only.
- `lightning_module.py` — `_forward_lr` / `_forward_sr` keep every constraint including the
  unclamped-loss-target rule and the `center_crop` zero-pad failure.
- `gan_module.py`, `datamodule.py` — the setup-hook ordering rule and the DistributedSampler
  repeat-padding hazard both kept in full.

**deleted as restatement/history (2).**

- `probe.py`'s account of what its two guarantees looked like *before* this module existed.
  The guarantees themselves are untouched.
- `metadata.py`'s restatement of two field lists that both builders' own `Returns:` already
  carry.

**moved (0).** Nothing needed a new owner.

**kept — unsure (0).**

## How "no rule was lost" was checked

Not by eye. **30 named constraints and failure modes** were listed from the pre-change text and
re-checked against the rewritten package: the two step axes, `DistributedSampler`'s
repeat-padding, `auto_insert_metric_name`'s silent directory-tree failure, the unclamped loss
target, `_RollingSaveMixin`'s shadowed-`super()` hazard, the pickle-clone "never first to open"
guarantee, `PERCEPTUAL_METRICS` as the direction source, and the rest. **30/30 present.**

Separately, every `__init__` parameter of the three touched public classes is still documented —
checked by walking the AST, not by reading.

Trivial `Args:`/`Returns:` were dropped from **private** methods only. `conventions` mandates
Google-style structure on public API and explicitly mandates nothing for private docstrings.

## Not yet at the gate, and here is the arithmetic

The target is aggregate **<= 1.00**. This lands at 1.67. Measured floor analysis over all 48
modules, so the remaining work is a known quantity rather than a guess:

| scenario | prose lines | ratio |
|---|---:|---:|
| today | 4,952 | 1.67 |
| every private docstring to one line | 3,371 | 1.14 |
| that, plus 15% off public docstrings | 2,953 | **0.996** |
| that, plus 30% off public docstrings | 2,534 | 0.85 |

**<= 1.00 is reachable without stripping a rule.** The calibration in `samples/` already
demonstrated 58% off a single docstring; the gate needs ~15% across public ones on top of a real
private-docstring pass. The remaining packages follow in later PRs.

`ruff` and `mypy` clean across all 48 modules.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -42,15 +42,9 @@ def _sr(pl_module: lightning.LightningModule) -> "SRLightning":
     with a narrower parameter type is unsound and rejected, so the narrowing
     happens at the boundary instead, once, with a name.
 
-    A cast rather than an ``isinstance`` check on purpose: a module without those
-    attributes already fails at exactly the same call today, and turning that into
-    a different failure is a behaviour change this seam has no business making.
-
-    Args:
-        pl_module: The module Lightning handed the hook.
-
-    Returns:
-        The same object, typed as what it has to be.
+    A cast, not an ``isinstance`` check: a module without those attributes
+    already fails at the same call today, and turning that into a different
+    failure is a behaviour change this seam has no business making.
     """
     return cast("SRLightning", pl_module)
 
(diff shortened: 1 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #212, opened 2026-09-01
chore: trim .gitignore to the tools this project actually uses

YeapJieShen:
Phase 1's packaging audit. `pyproject.toml` and `.gitignore` had never been examined, unlike the
rest of the tree.

## `pyproject.toml` — no change, and that is the finding

Audited and left alone. It is already tight at 143 lines, every dependency carries its
rationale inline as the project requires, and the checks that could have failed all pass:

- all four `docs/configuration.md` references resolve (that file is tracked)
- `requires-python = ">=3.12"`, `target-version = "py312"` and the `3.12`/`3.13` classifiers
  agree with the CI matrix
- the `[export]` and `[perceptual]` extras match their mypy overrides
- `filterwarnings = ["error", ...]` with two justified, scoped exemptions

## `.gitignore` — 213 -> 110 lines

It was the stock GitHub Python template with ~16 project-specific lines on top, carrying rules
for **Django, Flask, Scrapy, Celery, SageMath, Abstra, Cursor, PyInstaller, PyBuilder, Sphinx,
mkdocs, Spyder, Rope, IPython, Cython, Pyre, pytype and PEP 582** — none of which this project
uses — plus lockfile commentary for four package managers it does not use either.

Two rules got attention rather than deletion:

- **`*.log` was filed under `# Django stuff:`** and is genuinely wanted here, since a training
  run writes logs beside the checkout. Kept, and moved into this project's own section with a
  reason attached.
- **`lib/` and `lib64/`** are virtualenv-layout leftovers: dead (`.venv` is ignored wholesale)
  and broad enough to swallow a real source directory named `lib`. Dropped.

The project's own rules also gained the reason each exists, which none of them had.

## Proven inert, not asserted

The risk in trimming an ignore file is silently changing what is ignored. So:

- the full working tree's **ignored-path set is byte-identical** before and after
- **nothing became newly tracked**

Both checked by diffing `git status --porcelain --ignored` across the change.

Anything removed can come back — with the tool that needs it. The header says so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,47 @@
-# Extra files to ignore
+# ---------------------------------------------------------------------------
+# This project's own rules. Everything below the divider is the stock GitHub
+# Python template, trimmed to the tools this repo actually uses -- see the
+# note there before adding anything back.
+# ---------------------------------------------------------------------------
+
+# Datasets and derived caches: large, externally sourced, never committed.
 data
 .lmdb_cache/
+
+# Training runs: weights, event logs, metrics. Numbers that matter are
+# transcribed into the notes before a run is pruned.
 experiments/
+
+# Run configs are per-experiment and untracked; only the *.template.yaml
+# starting points ship. The untracked ones are mirrored outside the repo.
 templates/*
 !templates/*template*
 
-# Claude's local working notes (codebase index, triage, session history)
+# Claude's local working notes (codebase index, triage, session history).
 .claude/
 
 # LLM agentic coding artifacts (brainstorming specs, implementation plans).
 docs/superpowers/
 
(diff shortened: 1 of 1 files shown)


YeapJieShen:
Added the upstream attribution to the divider comment, as requested:

```
# Modified from the stock GitHub Python template:
#   https://github.com/github/gitignore/blob/main/Python.gitignore
```

It sits at the head of the trimmed half rather than the top of the file, so it is next to the list of what was removed — a reader can follow the link and diff against it directly.

Comment-only change; the ignored-path set was re-verified as identical.

YeapJieShen (commented):
Human reviewed, all good.

PR #213, opened 2026-09-01
refactor: compact prose in sisr/utils, losing no rule

YeapJieShen:
Second PR of the prose stack. Stacked on #211.

`sisr/utils` **1.47 -> 1.34 : 1**. Nine blocks across `cache.py` and `imresize.py`.

## Constraint ledger

**compressed (9).**

- `imresize.py`'s module docstring loses its narrative about which downstream lineage rounds
  when, and gains a flat list of the **three details load-bearing for byte-equality** — float64
  throughout, round-half-away-from-zero, and the inter-pass uint8 re-quantisation — each stated
  with the score its broken variant produced (19/57 and 0/57). Nothing about them is new; they
  were previously spread across three paragraphs.
- `cache.py`'s locking contract becomes **three numbered rules** where it was a twenty-line
  paragraph: a provably live holder is never preempted, takeover needs a dead pid **and** a
  stale sentinel, and waiting on a live holder is itself capped at `3 * lock_timeout`.
- `_acquire_lock` stops restating those three and keeps only what bites at the implementation —
  that "live" excludes *our own pid*, because a crashed builder's can be recycled to us, which
  on Windows is real rather than theoretical.
- `_publish`, `_try_load`, `_sibling_tmp_path`, `_start_heartbeat`, `get_buffer`,
  `_pid_alive_windows` — each keeps its failure mode and loses its narration.

**deleted (0). moved (0). kept — unsure (0).**

## Licences untouched

`imresize.py` carries the vendored MIT licence and `ssim.py` (next PR) the daala BSD one. Both
are **verbatim and unmodified** — checked, not assumed.

`ruff check`, `ruff format --check` and `mypy` clean.

### DIFF
--- a/sisr/utils/cache.py
+++ b/sisr/utils/cache.py
@@ -63,19 +63,11 @@ _kernel32 = ctypes.WinDLL("kernel32", use_last_error=True) if sys.platform == "w
 def _pid_alive_windows(pid: int) -> bool:
     """Windows liveness check for :func:`_pid_alive`, split out for testing.
 
-    ``OpenProcess`` failing does not always mean "no such process": a
-    builder running under another user account or session is a real,
-    live process this one simply lacks permission to query
-    (``ERROR_ACCESS_DENIED``) and must not be treated as dead. Only an
-    actually-missing pid reports something else (``ERROR_INVALID_PARAMETER``
-    in practice).
-
-    Args:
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #214, opened 2026-09-01
refactor: compact prose in sisr/metrics and sisr/datasets

YeapJieShen:
Third PR of the prose stack. Stacked on #213.

`sisr/metrics` 3.39 -> 3.37, `sisr/datasets` 1.92 -> 1.88, plus the two densest docstrings
outside any package pass (`cli.py::_parse_ckpt_path`, `losses/pixel.py::TotalVariationLoss`).

## `sisr/metrics` barely moved, and that is the finding

122 lines of code against 414 of prose — the worst ratio in the package — **because the prose is
the deliverable there.** It carries Ledig's own quoted sentence, the four load-bearing deviations
from `torchmetrics`, and a parity bound that is measured rather than padded. Cutting it would
delete the evidence that makes this project's SSIM claim checkable, which is the opposite of the
point. `ssim.py`'s vectorisation and float64 arguments become two headed claims instead of two
paragraphs; nothing else was safe to touch.

## Constraint ledger

**compressed (6).**

- `ssim.py` — the exact-vectorisation argument and the float64/parity-bound argument, both
  keeping the measured bracket (`1e-12` for a correct implementation, `5e-8` for the faintest
  real defect seen) that stops someone tightening `rel=1e-9`.
- `datasets/srresnet.py`, `datasets/srcnn.py` — step-by-step narration out, three rules kept and
  now stated as rules: **only the decode is memoized, never the crop**; **the grid comes from
  exactly one place**; **the checksum keys only on the file manifest**.
- `cli.py::_parse_ckpt_path` — keeps both substitutions, the subcommand-parser requirement and
  both failure modes, but the jsonargparse internals collapse to one paragraph and the
  consequence gets stated on its own, so *"re-inlining this as `self.parser` reintroduces
  exactly that"* is no longer buried at the end of twenty lines.
- `losses/pixel.py::TotalVariationLoss` — both traps kept and given headings: `target` is
  ignored, and the paper's `2e-8` presumes a `[-1, 1]` range so a `[0, 1]` processor halves the
  effective weight.

**deleted (0). moved (0). kept — unsure (0).**

The daala BSD licence in `ssim.py` is verbatim and unmodified — checked.

`ruff check`, `ruff format --check` and `mypy` clean across all 48 modules.

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -291,26 +291,22 @@ class SRLightningCLI(LightningCLI):
         (any nested subclass field with a fallback, e.g. ``eval_config``) — see
         below for which and why.
 
-        The reason: jsonargparse's subclass-merge machinery
-        (``ActionTypeHint._check_type``) looks up the field's *previous* value as
-        ``cfg.get(self.dest)`` — ``self.dest`` is the action's bare name (``"model"``),
-        not a subcommand-qualified one. Reached through the subcommand's own
-        parser, ``cfg`` is already scoped to that subcommand, so ``cfg.get("model")``
-        finds the real, already-resolved value and merges our override's keys onto
-        it, preserving every sibling key we don't mention (``model``, ``processor``,
-        nested ``class_path`` identities) and any subclass-only default our override
-        never names. Reached through the top-level parser, ``cfg`` is the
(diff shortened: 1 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #216, opened 2026-09-01
docs: split the reproduction reference out of the README

### LINKED ISSUE #198: Split the reproduction reference out of the README
## What changed

The README is **567 lines across 13 top-level sections**, up from 411 before the documentation
stack landed (#185, #186, #188 added the fidelity tables, the paper-silent-parameters note and
the weight-licence policy). That is +38% in one batch, and the growth is not evenly spread:

| section | lines |
|---|---|
| Comparability | 106 |
| Reproduction results | 102 |
| SRGAN | 81 |
| Losses | 66 |
| Config overrides and subclass defaults | 59 |

## The problem is not length, it is that one file has several audiences

The README currently answers at least four unrelated questions:

1. **"How do I run this?"** — Install, Quickstart, ONNX export, Development.
2. **"How is it built?"** — How it works, Models, Losses, SRGAN.
3. **"Do the numbers reproduce, and under exactly what convention?"** — Reproduction results,
   Comparability.
4. **"What am I allowed to do with the weights?"** — Model weights, License.

Group 3 is the one that keeps growing, and it is *reference material a reader consults with a
specific question*, not narrative anyone reads top to bottom. It is also the group most likely to
grow again: every new architecture adds rows, and every convention question adds a paragraph
explaining which numbers are comparable to which.

A reader arriving to run the thing now scrolls past ~200 lines of metric-convention argument to
reach ONNX export.

## Proposal

Split the reference material out, keep the narrative in the README:

- `docs/reproduction.md` — Reproduction results + Comparability (the fidelity tables, the
  colorspace/SSIM convention rules, what is and is not comparable to the papers).
- README keeps Install, Quickstart, How it works, Models, Losses, SRGAN, config overrides, ONNX,
  Development, weights, licence — and links to the reference doc from a short summary with the
  headline figures only.

The constraint that matters: **each fact keeps exactly one owner.** The README should cite the
headline numbers and link, never restate the tables, or the two copies will drift and the
convention caveats will end up attached to only one of them.

## Note on history

This was raised and deferred once before, around 2026-08-26, when the README was smaller and the
tables were not yet published. The documentation stack has now landed the material that made it
worth deferring, so it is worth re-deciding rather than assuming the earlier deferral still holds.

## Not urgent

Nothing is broken and no number is wrong. This is a navigability question, filed so the next
addition to the fidelity tables is a decision about where it goes rather than another append.

YeapJieShen:
Closes #198. Stacked on #208, because it moves the exact sections #208 edits.

**README 629 -> 384 lines.** `Reproduction results` and `Comparability` — 269 lines of tables,
convention rules and artifact provenance — move whole into **`docs/reproduction.md`**, which
becomes their single owner.

## Why these two sections and not others

#198's reasoning holds: the README answers four unrelated questions, and this is the group that
is *reference material a reader consults with a specific question* rather than narrative anyone
reads top to bottom. It is also the group that keeps growing — every architecture adds rows,
every convention question adds a paragraph. A reader arriving to run the thing was scrolling
past ~200 lines of metric-convention argument to reach ONNX export.

## The constraint that mattered, and how it was checked

#198 is explicit: **each fact keeps exactly one owner.** The README must cite and link, never
restate, or the two copies drift and the convention caveats end up attached to only one.

So this was verified mechanically rather than by eye:

- **all 269 moved lines are present** in `docs/reproduction.md` — nothing lost
- **zero lines of the moved block remain in the README** — no second owner

The first draft failed the second check on one sentence, which was reworded rather than left.

## What the README keeps

A short summary carrying only the headline *shape* — one architecture reproduces and one does
not, SRCNN has no publishable row, one papers-silent row remains open, and a figure is
comparable only to one computed the same way — then a link. No table, no figure the reference
doc owns.

## Links

- the one anchor from outside the moved range (`Model weights` -> `Comparability`) is repointed
  at `docs/reproduction.md#comparability`
- anchors inside the moved block were already intra-section and still resolve
- `CONTRIBUTING.md`'s link to `README.md#install` is unaffected — `Install` stayed

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,282 +236,27 @@ first validation rather than at step 0; read the comment beside its
 
 ## Reproduction results
 
-Both architectures are scored against the source paper's own tables, under the source
-paper's own metric convention. **One reproduces and one does not**, and both are published
-here — a table that shows only the win is not evidence of anything.
+**One architecture reproduces its paper and one does not.** Both results are published, because
+a table that shows only the win is not evidence of anything.
 
-### SRResNet reproduces Ledig et al.
+- **SRResNet reproduces Ledig et al.** Within **~0.1 dB** on Set5/Set14/BSD100 PSNR-Y, on
+  DIV2K-800 against their 350k ImageNet images, with a small *uniform* SSIM residual.
+- **SRGAN-VGG54 does not**, and the cause is isolated to a single constant: the paper's own
+  stated purpose for its `feature_scale` is measurably not achieved on this data.
+- **SRCNN has no publishable row** — every SRCNN run predates the current degradation and
+  metric conventions, so none are comparable to anything, including each other.
+- **One place the papers stop specifying** remains open; three were closed on 2026-09-01 from
+  the authors' own released code, which also surfaced a **2.4x disagreement between that code
+  and the SRCNN paper's stated training budget**.
 
-Ledig et al., ["Photo-Realistic Single Image Super-Resolution Using a Generative Adversarial
-Network"](https://arxiv.org/abs/1609.04802), Table 2, SRResNet row, ×4. Ours is trained on
-DIV2K-800; theirs on 350k ImageNet images. Same 1e6 iterations, same 1e-4 learning rate.
+**🚨 A figure is comparable only to one computed the same way.** The SSIM convention, the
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #217, opened 2026-09-01
fix: derive the evaluation border from scale instead of pinning a constant

YeapJieShen:
First half of #207, per the decision on that issue. **`crop_border` derives from `scale`.**

## Your read was right: we already had this convention

`sisr/training/config.py` already documented it — *"Standard SR-evaluation convention is to crop
the outer `scale` pixels"* — and the SRCNN authors' released demo code does exactly that
(`shave(..., [up_scale, up_scale])`). Both shipped architectures already satisfied it:

| | scale | crop_border | |
|---|---:|---:|---|
| SRCNN | 3 | 3 | equal, by coincidence |
| SRResNet / SRGAN | 4 | 4 | equal, by coincidence |

**The rule was written down and nothing enforced it.** The constants are right only because we
happen to ship ×3 and ×4. A ×2 config silently kept 3 or 4 — and a border changes the reported
PSNR/SSIM without changing the model.

## What changed

`crop_border` becomes `int | None`. **`None` is the sentinel for "the field convention: the
model's scale"**, resolved once by `SRLightning` at construction — before the scorer is built, so
the scorer and every artifact's `sisr_meta` record the number actually used. An explicit int
always wins, `0` included. The base default stays `0`, so bare `SREvalConfig()` is unchanged and
the 60-odd existing uses are untouched.

**A side effect worth having:** the subclass defaults are now the same sentinel as the base, so
the documented `--ckpt_path` reversion of subclass defaults to base **can no longer change the
border**. One less thing that reversion can silently alter.

## Evidence

**RED → GREEN, proven on pre-fix `main`:**

```
E   AssertionError: a scale-2 run must crop 2 border pixels, not the shipped constant
E   assert 4 == 2
```

`test_subclass_crop_border_follows_the_models_scale_not_a_constant` passes after. A second test
pins that an explicit `crop_border=8` is never overwritten by the derivation.

**mypy caught the real consequence** rather than me: the scorer could now receive an unresolved
`None`. That is a **loud `ValueError` naming the cause**, never a silent crop of 0 — an
unresolved sentinel means the config never passed through `SRLightning`, which is a bug, not a
default.

**No shipped number changes**: 3 and 4 are exactly what ×3 and ×4 now derive.

376 tests pass locally across `metrics`, `models` and `training`; `ruff check`,
`ruff format --check` and `mypy` clean.

## Not in this PR

The other half of #207 — adopting the authors' `same`+`replicate` inference path with an opt-out
to keep valid convolutions at both train and test — changes what SRCNN *outputs* at eval time and
needs its own tests. It follows separately.

### DIFF
--- a/sisr/metrics/scoring.py
+++ b/sisr/metrics/scoring.py
@@ -128,8 +128,20 @@ class SRScorer:
 
         Returns:
             The cropped tensors, in the order given.
+
+        Raises:
+            ValueError: If ``crop_border`` is still ``None``. That means the
+                config never passed through ``SRLightning``, which resolves the
+                derive-from-scale sentinel; scoring against an unresolved border
+                would silently pick a different region than the run intends.
         """
         n = self.eval_config.crop_border
+        if n is None:
+            raise ValueError(
+                "eval_config.crop_border is still None at scoring time. It is a sentinel "
+                "meaning 'crop the model's scale', resolved by SRLightning at construction "
+                "-- build the scorer from a module's eval_config, or set an explicit int."
+            )
         if n <= 0:
             return tensors
         return tuple(t[..., n:-n, n:-n] for t in tensors)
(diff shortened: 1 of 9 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #219, opened 2026-09-01
fix: give SRCNN biases the authors learning rate

### LINKED ISSUE #210: SRCNN biases train at 10x the authors' learning rate
Found while settling #202 against the SRCNN authors' released Caffe files (`SRCNN_train.zip`,
retrieved 2026-09-01, SHA-256
`001146419f7acfb12a3e7929c8acd5de88a08d687d6881085f81321ad6982b1a`).

## What the authors specify

Every convolution layer in `SRCNN_net.prototxt` carries two `param` blocks. In Caffe these
apply to the layer's blobs in order — **first weights, then bias**:

| layer | weight `lr_mult` | bias `lr_mult` |
|---|---:|---:|
| conv1 | `1` | `0.1` |
| conv2 | `1` | `0.1` |
| conv3 | `0.1` | `0.1` |

With `base_lr: 0.0001`, that is:

| | weight LR | bias LR |
|---|---:|---:|
| conv1 | 1e-4 | **1e-5** |
| conv2 | 1e-4 | **1e-5** |
| conv3 | 1e-5 | 1e-5 |

**Every bias trains at one tenth of the base rate**, independently of the per-layer weight
schedule.

## What we ship

`layer_lrs: [1e-4, 1e-4, 1e-5]`, and the param-group builder assigns one LR to *all* of a
layer's parameters:

```python
param_groups = [
    {"params": list(layer.parameters()), "lr": lr}
    for layer, lr in zip(conv_layers, lrs, strict=False)
]
```

`list(layer.parameters())` is weight **and** bias. So conv1 and conv2 biases train at `1e-4` —
**10x the authors' rate**. conv3 happens to agree, because its weight rate is already 1e-5.

The per-layer *weight* schedule matches the authors exactly, and so does `init_std: 0.001`;
this is the one part of the port that does not.

## Why it is not fixed in the PR that found it

`layer_lrs` is per-`Conv2d` and cannot express a weight/bias split, so this needs a schema
change rather than a value change, and there is more than one defensible surface:

1. **`bias_lr_mult: float | None`** on `SRCNNTrainingConfig` — smallest change, reads exactly
   like the prototxt, but adds a knob that only one architecture uses.
2. **Per-parameter-group LRs** — `layer_lrs` becomes a list of `{weight, bias}` pairs. More
   expressive, and a breaking change to a published config key.
3. **Adopt Caffe's model wholesale** — a base LR plus per-blob multipliers. Most faithful to
   the source, largest departure from how the rest of this project configures optimizers.

## Decide alongside #207

Both are "how far do we follow the SRCNN authors' released code". Answering them separately
risks adopting their optimizer schedule while keeping our evaluation geometry, or the reverse
— a half-reproduction that is harder to reason about than either whole.

## Blast radius

**None published.** SRCNN has no publishable row.

How much it matters in practice is unmeasured: the affected parameters are 96 bias scalars out
of ~57k, all initialised to zero. Small, but it is a fidelity claim, and the point of this
project is that the fidelity claims are checkable.

YeapJieShen:
## What changed

SRCNN's conv1 and conv2 biases now train at `1e-5`, the rate the authors specify, instead of
`1e-4`. A `layer_lrs` entry may be a `[weight_lr, bias_lr]` pair as well as a float.

## Why

The authors' released `SRCNN_net.prototxt` gives every convolution two `param` blocks, which
Caffe applies to the layer's blobs in order — weights, then bias:

| layer | weight `lr_mult` | bias `lr_mult` | weight LR | bias LR |
|---|---:|---:|---:|---:|
| conv1 | `1` | `0.1` | 1e-4 | **1e-5** |
| conv2 | `1` | `0.1` | 1e-4 | **1e-5** |
| conv3 | `0.1` | `0.1` | 1e-5 | 1e-5 |

Every bias trains at a tenth of `base_lr`, independently of the per-layer weight schedule.
`layer_lrs` assigned one LR to `list(layer.parameters())` — weight *and* bias — so conv1 and
conv2 biases ran at 10x. conv3 happened to agree, because its weight rate is already 1e-5.

The per-layer weight schedule and `init_std: 0.001` already matched; this was the one part of
the port that did not.

## Why a pair, and not a `bias_lr_mult`

A single multiplier would have had to say what it multiplies. These LRs are **absolute** and
the optimizer's own `lr` is ignored when they are set, so there is no base rate in scope:
relative-to-layer-LR gives conv3 a bias rate of 1e-6, not the authors' 1e-5. Absolute pairs
read straight off the table above with no inference step, and a bare float still means "weight
and bias together", so nothing existing changes meaning.

## Evidence

Four tests, all confirmed RED on `main` before the fix:

- `test_configure_optimizers_per_blob_lrs` — weights `[1e-4, 1e-4, 1e-5]`, biases
  `[1e-5, 1e-5, 1e-5]`, read back off the built optimizer's param groups.
- `test_configure_optimizers_scalar_and_pair_entries_mix` — a bare float still means both.
- `test_configure_optimizers_per_blob_wrong_pair_length_raises`
- `test_configure_optimizers_per_blob_bias_lr_without_bias_raises` — a bias LR aimed at a
  `bias=False` layer is a config error, not a silent drop.

The CLI test resolves the shipped template and asserts the nested value, which is also what
proves jsonargparse handles the `float | list[float]` union.

## Blast radius

None published — SRCNN has no publishable row. 96 bias scalars out of ~57k, all initialised to
zero; how much it matters in practice is unmeasured. It is a fidelity claim, and the point of
this project is that the fidelity claims are checkable.

## Merge

**Rebase, not squash** — this PR carries a code commit and a docs commit, and per
CONTRIBUTING.md those stay separate.

Closes #210

### DIFF
--- a/docs/reproduction.md
+++ b/docs/reproduction.md
@@ -113,8 +113,17 @@ both rows at once:
 | `hdf5_data_param.batch_size` (train) | `128` | `batch_size: 128` (was `64`) |
 | `max_iter` | `15000000` | see below |
 | `base_lr` / `momentum` / `weight_decay` | `0.0001` / `0.9` / `0` | already matched |
-| per-layer `lr_mult` (weights) | `1`, `1`, `0.1` | already matched — `layer_lrs: [1e-4, 1e-4, 1e-5]` |
+| per-layer `lr_mult` (weights) | `1`, `1`, `0.1` | already matched |
+| per-layer `lr_mult` (biases) | `0.1`, `0.1`, `0.1` | **now matched** — see below |
 | `weight_filler` std | `0.001` | already matched |
+
+The **bias** row was not matched, and the same file is what exposed it. Caffe applies a
+layer's two `param` blocks to its blobs in order — weights, then bias — so conv1 and conv2 pair
+`lr_mult` 1 with 0.1 and every bias trains at a tenth of `base_lr`, independently of the
+per-layer weight schedule. This project assigned one rate to a layer's weight *and* bias, so
+conv1 and conv2 biases trained at 10x the authors' rate; conv3 happened to agree, because its
+weight rate is already 1e-5. A `layer_lrs` entry may now be a `[weight_lr, bias_lr]` pair, and
+the shipped SRCNN recipe uses one per layer.
 
 `max_iter` settles one thing and opens another.
 
--- a/sisr/models/srcnn/config.py
+++ b/sisr/models/srcnn/config.py
@@ -17,6 +17,16 @@ from sisr.processors.base import SRProcessor
 from sisr.training.config import SREvalConfig, SRTrainingConfig
 
 
+def _authors_layer_lrs() -> list[float | list[float]]:
+    """``[weight_lr, bias_lr]`` per conv, from ``base_lr`` x the prototxt's ``lr_mult``.
+
+    A function rather than a lambda so the element type is ``float |
+    list[float]`` -- ``list`` is invariant, so an inferred
+    ``list[list[float]]`` will not satisfy the field.
+    """
+    return [[1.0e-4, 1.0e-5], [1.0e-4, 1.0e-5], [1.0e-5, 1.0e-5]]
+
+
 @dataclass
 class SRCNNTrainingConfig(SRTrainingConfig):
     """SRCNN-paper-faithful training defaults.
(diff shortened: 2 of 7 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #220, opened 2026-09-01
feat: run SRCNN inference the way its authors do

### LINKED ISSUE #207: SRCNN inference padding and evaluation border both differ from the authors' demo code

YeapJieShen:
## What changed

SRCNN now runs inference with SAME padding and replicated borders, as the authors' released
demo code does, while still training with `pad: 0` as they do. `SRCNN` gains `eval_padding`
and `eval_padding_mode`; the shipped template sets them.

## Why

`SRCNN.m` convolves with `imfilter(..., 'same', 'replicate')` and `demo_SR.m` then shaves
`scale` pixels per side before PSNR. This project convolved **valid** at inference too, losing
6 px per side before `crop_border` shaved any, so at x3 it scored:

| | region scored on an HxW ground truth |
|---|---|
| authors | rows `4 .. H-3` |
| before this PR | rows `10 .. H-9` |

Strictly inside theirs, and not the same measurement — on a metric where border handling is
known to matter.

The two divergences were entangled: `crop_border` being a constant where the authors' border is
`scale` (fixed already, #217) and the valid-conv loss underneath it. Fixing either alone moves
*away* from the authors. With a full-size output the center-crop that aligns SR to HR becomes a
no-op, so the border alone decides the region — and it is now the authors' region.

## The trade, decided rather than assumed

Same-padded inference scores pixels whose receptive field includes invented (replicated)
content. Valid convolution never does, and is arguably the more honest measurement — but then
the numbers are not comparable to the paper's. Reproducing the paper won. **The opt-out is
deleting the two `eval_padding` lines from the config**, which restores valid convolution at
train and test.

## Two things that are easy to get wrong

- **The border is re-derived at every convolution**, matching the authors' per-layer
  `imfilter`. Pre-padding the input once by the full 6 px receptive field and convolving valid
  is the cheap-looking equivalent and a *different function* — later layers would see a border
  built from replicated input pixels rather than replicated features.
  `test_eval_padding_pads_per_layer_not_the_input_once` pins the difference.
- **Mutating a built `Conv2d`'s `.padding` / `.padding_mode` does not work.** `__init__`
  precomputes the `'same'` pad tuple, so a mutated module silently keeps convolving valid —
  it would have claimed the authors' path while doing the old one. Hence an explicit
  pad-then-convolve.

`eval_padding` applies only outside training mode, so **the model's output shape now depends on
training mode**. That is deliberate — it is what the authors' own setup is — and it is written
into the `forward` docstring.

## Evidence

Tests, all confirmed RED before the change (the constructor did not accept the argument):

- `test_eval_padding_only_applies_out_of_training_mode` — 33x33 in gives 21x21 training,
  33x33 eval.
- `test_eval_padding_none_keeps_valid_in_both_modes` — the opt-out, and the default.
- `test_eval_padding_pads_per_layer_not_the_input_once`
- `test_eval_padding_matches_an_explicit_per_layer_reference` — exact arithmetic against a
  hand-written pad-then-valid-conv loop.
- `test_eval_padding_rejects_an_even_kernel`, `test_eval_padding_is_recorded_in_hparams`
- `test_eval_padding_makes_the_scored_region_the_authors` — through `SRLightning`: the SR
  field is full-size, so `center_crop` takes nothing and `crop_border` (3, derived from
  `scale`) is the whole difference.

Metadata tests updated: the two new hparams are architectural and must round-trip through
checkpoints and the ONNX `sisr_meta` props. The ONNX export now emits a constant-folding
notice — `F.pad(mode='replicate')` lowers to an `onnx::Slice` the folder declines to fold. It
is an optimization notice, not a correctness one; the graph exports, and the test filters it by
message rather than broadening the suite's `error` policy.

## Blast radius

None published — SRCNN has no publishable row, so no existing figure is invalidated. SRCNN's
validation metric during training does change, which is the intended effect.

## Merge

**Rebase, not squash** — code commit and docs commit stay separate.

Stacked on #219. Closes #207

### DIFF
--- a/docs/reproduction.md
+++ b/docs/reproduction.md
@@ -148,15 +148,33 @@ unsourced number.
 No published number changes: SRCNN has no publishable row.
 
 **The evaluation border also left**, for the opposite reason: it turned out to be specified
-after all, and we differ from it. The authors' demo package (`SRCNN_v1.zip`, retrieved
+after all, and we differed from it. The authors' demo package (`SRCNN_v1.zip`, retrieved
 2026-09-01, SHA-256 `bfa68ca613c1326a59e0c34353205a254ab2b67e34df7f04e28eef567980af30`)
 shaves **`scale` pixels per side** before computing PSNR — not a constant — and runs the
 network with **`same` padding and replicated borders** at test time, though it trains with
-`pad: 0`. This project uses `padding: valid` at inference and then `crop_border: 3`, so at ×3
-we score a region strictly *inside* theirs, and our constant agrees with `scale` only at ×3.
-The two differences are entangled — changing the border alone moves us further away, not
-closer — so the choice between reproducing their measurement and keeping a path that never
-scores invented border pixels is tracked as its own decision, not settled here (#207).
+`pad: 0`. This project used `padding: valid` at inference and then `crop_border: 3`, so at ×3
+it scored a region strictly *inside* theirs, and the constant agreed with `scale` only at ×3.
+
+The two differences were entangled — changing the border alone moves *away* from the authors,
+not toward them, because the valid-convolution loss stays underneath — so both are now
+adopted together:
+
+| | authors | ships now |
+|---|---|---|
+| inference padding | `same`, replicated border | `eval_padding: same`, `eval_padding_mode: replicate` |
+| training padding | `pad: 0` | `padding: 0`, unchanged |
+| border shaved before PSNR | `scale` px/side | `crop_border` derived from `scale` |
+
+The replicated border is re-derived at **every** convolution, as the authors' per-layer
+`imfilter` does; pre-padding the input once and convolving valid is a different function.
+`eval_padding` applies only outside training mode, so the model is deliberately not the same
+function in the two modes — which is what the authors' own setup is.
+
+The trade this makes is real and was decided rather than assumed: same-padded inference scores
+pixels whose receptive field includes invented (replicated) content. Keeping valid convolution
+never does, and is arguably more honest, but then the numbers are not comparable to the paper's
+on a metric where border handling is known to matter. Reproducing the paper won. Deleting the
+two `eval_padding` lines from the config restores valid convolution at train and test.
 
 #### What did *not* settle: SRResNet's PReLU
 
--- a/sisr/models/srcnn/model.py
+++ b/sisr/models/srcnn/model.py
@@ -25,6 +25,14 @@ class SRCNN(SRModel):
             original architecture).
         padding: ``'valid'``, ``'same'``, or an explicit pixel count.
             Defaults to ``'valid'``.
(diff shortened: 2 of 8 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #223, opened 2026-09-01
feat: record the host's power state, and say so before a long run

### LINKED ISSUE #201: Training runs do not check mains power, so a contaminated rate looks sound
## The gap

The rate harness refuses to measure unless the host is on mains under the Best-performance
overlay — it reads the registry overlay value, because `powercfg /getactivescheme` reports the
underlying scheme and `nvidia-smi` cannot see the enforced power limit under WSL.

A training run has no such guard. `fit` starts happily on battery.

## What that cost, measured

A long adversarial run was started while the power supply was switched off by accident. It ran at
**172 batches/min**, died when the battery ran flat 87 minutes in, and — the part that matters —
**every timing taken from it went into the notes as if it were sound.** Re-measured after
restoring mains, over 5,000 batches under a verified check: **366 batches/min**. A 2.13x
difference, large enough that the run's projected length was wrong by more than a factor of two,
and the wrong figure was already written down.

The crash was loud and cost 87 minutes. The bad number was silent and would have outlived it.

## Proposal

Assert the same condition at the start of `fit`, and make it **advisory rather than fatal**:

- Warn loudly, naming the overlay actually in force, and record the power state in the run's own
  metadata alongside the other provenance the artifacts already carry.
- Do not refuse. A correctness-only run on a laptop in a cafe is legitimate; what is not
  legitimate is *not knowing* afterwards which conditions a number came from.

Recording it in metadata is the half that pays off later: a rate quoted from a run whose
provenance says "battery" is then self-refuting rather than plausible.

## Notes

- Cheap: the check already exists and takes one PowerShell call, once, at startup.
- WSL/Windows-specific. On another platform it should degrade to a no-op, not an error.
- Related to the standing provenance bar — a throughput number without its conditions is a lead,
  not a result.

## Acceptance

- [ ] `fit` warns when not on mains under Best performance, naming what is in force
- [ ] The power state appears in run metadata
- [ ] No-ops cleanly where the check does not apply

YeapJieShen:
## What changed

`fit` now warns when the host is not in the power condition throughput figures assume, and every
artifact the run writes records that state.

## Why

A rate measured off mains is worth about half. The same adversarial run, same config:

| condition | rate |
|---|---:|
| draining battery | 172 batches/min |
| mains, Best-performance overlay | 366 batches/min |

It died when the battery ran flat 87 minutes in — and **every timing taken from it went into the
notes as if it were sound**. The crash was loud and cost 87 minutes; the bad number was silent
and would have outlived it. The rate harness already refuses to measure without this check.
`fit` had no guard at all.

## Advisory, never fatal

A correctness-only run on a laptop is legitimate. What is not legitimate is not knowing
afterwards which conditions a number came from. So this warns, records, and refuses nothing.

The metadata is the half that pays off later: a rate quoted from a run whose provenance says
`BATTERY` is self-refuting rather than merely plausible.

## Three decisions, not accidents

- **Read once per process, not per save.** The warning and the metadata want the *same* answer,
  and a per-save read would put a subprocess call — and a chance to hang — in the checkpoint
  path. The recorded field is therefore the state near the *start* of the run, and says so.
- **Silent when the state is unknown.** There is no instrument outside WSL/Windows, so warning
  there would fire on every run everywhere and say nothing. Unknown is a recorded state, not a
  suspicion. `fit`-only for the same reason: `validate`/`test`/`export` are short.
- **PowerShell is the instrument** because nothing else can see it: `nvidia-smi` cannot read the
  enforced power limit under WSL, and `powercfg /getactivescheme` reports the *underlying
  scheme* rather than the overlay sitting on top of it, so the registry value is what gets read.

`sisr/utils/power.py` is stdlib-only, keeping `sisr.utils` torch-free as its package docstring
requires — the guard test for that covers the package, not just `cache`.

## Evidence

13 tests in `tests/utils/test_power.py` covering both wirings and every degradation path:
mains + Best performance is the reference; battery is not; **mains under a different overlay is
not either** (the overlay is half the condition); a host reporting no battery counts as mains;
no `powershell.exe`, a raising probe, a timeout and unreadable output each degrade to unknown
rather than raising; the warning fires on battery, is silent on the reference condition, and is
silent on unknown; the state is read exactly once per process; `as_dict` is plain scalars, for
`weights_only` safety.

Plus `test_fit_warns_when_the_host_is_not_on_reference_power` and
`test_non_fit_subcommands_do_not_check_power` in the CLI tests, and
`test_build_metadata_records_the_hosts_power_state` for the artifact shape.

The suite gains one `filterwarnings` ignore, because no CI runner is a reference bench and the
warning is advisory by design. It is matched by message, and `pytest.warns` is unaffected by it,
so the tests that assert the warning still assert it.

`power` is additive metadata on the same terms as `kind` — `FORMAT_VERSION` is unchanged, and
its absence on an older file means *unrecorded*, not mains.

Stacked on #222. Closes #201

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,9 @@ filterwarnings = [
     "ignore::FutureWarning:lightning.*",
     # Post-optimization state is exactly what a checkpoint should hold.
     "ignore:Using ModelCheckpoint with manual optimization:UserWarning",
+    # Advisory by design, and no CI runner is a reference bench. The tests that
+    # care assert it with pytest.warns, which this does not suppress.
+    "ignore:Power state is not the reference condition:UserWarning",
 ]
 
 [tool.ruff]
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -41,6 +41,7 @@ from lightning.pytorch.cli import ArgsType, LightningArgumentParser, LightningCL
 
 from .export import to_onnx
 from .training import SRDataModule, SREvalConfig, SRLightning, SRTrainingConfig
+from .utils.power import warn_unless_reference_power
 
 # Checkpoints saved before SRLightning.__init__ stopped flattening self.hparams for
 # TensorBoard display (see its docstring) stored ONLY these two dataclasses' fields —
@@ -260,12 +261,20 @@ class SRLightningCLI(LightningCLI):
         )
 
     def before_instantiate_classes(self) -> None:
-        """Apply process-global flags (``matmul_precision``) before class instantiation."""
+        """Apply process-global flags, and check the host's power before a long run.
+
+        The power check is ``fit``-only and advisory: it never refuses, because a
+        correctness-only run on a laptop is legitimate. What is not legitimate is
+        not knowing afterwards which conditions a timing came from, so the state
+        is also recorded in every artifact the run writes.
+        """
         if self.subcommand is None:
             return
         precision = self.config[self.subcommand].matmul_precision
         if precision is not None:
             torch.set_float32_matmul_precision(precision)
+        if self.subcommand == "fit":
+            warn_unless_reference_power()
 
     def _parse_ckpt_path(self) -> None:
         """Reload ``--ckpt_path`` hyperparameters via ``_reconstruct_ckpt_hparams`` first.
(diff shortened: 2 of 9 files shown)


YeapJieShen:
Closing unmerged, at the maintainer's direction: this is host-specific and does not belong in a general-purpose framework.

The module hardcodes a Windows registry path, a Windows power-overlay GUID and a PowerShell invocation. Every other contributor and CI leg would carry that surface to get an "unknown" back, and the metadata field would be null in almost every artifact anyone else produced.

The need is real but the scope was wrong: it is a property of one measurement rig, so it stays in that rig's own harness, which already refuses to measure off mains. Nothing here is lost -- the check exists where it applies.

The fix in #222 is unaffected and stays open.

PR #225, opened 2026-09-01
refactor: compact the prose in the four packages that had none

YeapJieShen:
## What changed

Prose compaction across the four packages the earlier passes did not reach: root modules,
`losses/`, `models/`, `processors/`. **−137 prose lines, code unchanged** (2,990 → 2,991, one line
from a reflow).

| | before | after |
|---|---:|---:|
| prose : code, whole package | **1.639** | **1.593** |
| root modules | 1.74 | 1.52 |
| `models/` | 1.29 | 1.23 |
| `losses/` | 1.29 | 1.29 |
| `processors/` | 1.04 | 1.00 |

`sisr/models/srcnn/` is **deliberately untouched** — #219 and #220 both rewrite files in it, and a
prose diff on top would collide for no gain. It is the obvious next target once they land.

## Constraint ledger

Every removed block, classified. No constraint, failure mode, or `Args:`/`Returns:`/`Raises:`
entry was deleted; nothing was moved out of the codebase.

| file | cut | class |
|---|---|---|
| `colorspace.py` | `(torch.Tensor)` type prefixes in every `Args:`/`Returns:` | **deleted as restatement** — the signature already says it |
| `colorspace.py` | the studio-range PSNR offset derivation | **compressed to its conclusion**; the test that asserts the identity is named |
| `colorspace.py` | "these pure tensor functions live on their own so callers can…" | **compressed** — kept as one clause |
| `cli.py` | `_parse_ckpt_path`'s two-parser explanation | **compressed** — the *rule* (never `self.parser`), the mechanism (`cfg.get(self.dest)` is unqualified) and both failure modes all kept verbatim in substance |
| `cli.py` | `parse_arguments`' collision walk-through | **compressed**; the "catch, do not pre-scan" lesson kept as its own bolded rule |
| `cli.py` | `_CKPT_HPARAM_KEYS`' history of what old checkpoints stored | **deleted as narrative about what the code used to do** — the surviving rule is which keys are loadable |
| `cli.py` | `set_parsing_settings` rationale | **compressed**; the alternative (disabling the selector by name) kept |
| `artifacts.py` | "**Why it exists**" — four sentences on the pre-refactor shape | **deleted as narrative**; the one-owner rule kept |
| `artifacts.py` | "Why safetensors" | **compressed** — the consumer-exposure argument is the load-bearing half and stays |
| `export.py` | per-processor consequences, as prose | **compressed into a bulleted table**; the SRCNN Y→Y limitation kept, emphasised |
| `losses/base.py` | `last_terms` protocol paragraph | **compressed**; in-place-write rule and all three replacement triggers kept |
| `losses/pixel.py` | Charbonnier ε-convention paragraph | **compressed**; both live conventions kept |
| `losses/pixel.py` | TV isotropy + range-scaling traps | **compressed**; both traps kept as bullets, including the 2× weight consequence |
| `models/base.py` | `variant_tag` / `input_contract` "abstract not inferred" rationale | **compressed**; the rule and its reason kept |
| `processors/base.py` | `output_range` / `output_colorspace` rationale | **compressed**; the silent-wrong-image failure kept |
| `models/srresnet/config.py` | `init_strategy`, `scale` narrative | **compressed** |
| `models/srgan/config.py`, `processors/rgb.py` | `Args:` bodies restating the class docstring | **compressed** |

**Kept — unsure (0).** Nothing in this pass was a judgement call I could not settle; where a cut
felt close, the text was compressed rather than removed.

## One defect found, not moved

`SRResNetEvalConfig.crop_border` documented "Overrides the base default to `4`". The field has
been `None` since the border started deriving from `scale` (#217) — the docstring described a
value the code no longer holds. Corrected to state the sentinel and what it resolves to here.

## About the gate

**This does not reach the gate, and the gate's replacement number needs deciding with the
measurement in hand rather than before it.** `conventions.md` currently demands ≤ 1.00, which is
unreachable without deleting rules. The four packages in scope held 1,324 prose lines between
them; a pass that loses nothing took 10.4% of that, which is 137 lines and lands at **1.593**.

Reaching 1.40 needs 716 lines gone in total. The remaining 579 can only come from `training`,
`utils`, `metrics` and `datasets` — all already compacted once — so it is a *second* pass over
already-compacted code, not this one. That is a separate decision, and this PR does not presume
it: no gate value is changed here.

## Evidence

`ruff check`, `ruff format --check` and `mypy sisr/` all clean. Tests for every touched package
pass — no behaviour is in this diff, so CI's full run is the check that matters.

### DIFF
--- a/sisr/artifacts.py
+++ b/sisr/artifacts.py
@@ -1,26 +1,21 @@
 """The distributable weights artifact: one owner for writing, reading and checking it.
 
-A run produces two kinds of file. The resumable ``.ckpt`` belongs to Lightning —
-it carries optimizer moments, loop state and hyperparameters, and stays exactly
-as it is. This module owns the other one: the bare, optimizer-free weights file
-that gets handed to somebody else.
+A run produces two kinds of file. The resumable ``.ckpt`` is Lightning's, and
+stays as it is. This module owns the other: the bare, optimizer-free weights
+file handed to somebody else. Writing, reading and validation live here so the
+payload's shape has one owner rather than being re-derived per reader.
 
-**Why it exists.** The payload used to be written in one place and its shape
-re-derived by hand in three readers, one of which checked nothing at all. The
-format was `torch.save`, decided at the write site. Adding a second format meant
-editing four files, and the readers had no way to agree with each other.
+**Why safetensors.** Not because our readers are unsafe -- every ``torch.load``
+here passes ``weights_only=True`` against a pinned torch floor. The exposure is
+a *consumer* loading a published file without it, or on an older torch.
+safetensors removes that by construction rather than by mitigation, and it is
+what the surrounding ecosystem reads.
 
-**Why safetensors.** Not because our own readers were unsafe — every
-``torch.load`` here already passes ``weights_only=True`` and the torch floor is
-pinned for it. The exposure is a *consumer* loading a published file with that
-turned off, or on an older torch. safetensors removes it by construction rather
-than by mitigation, and it is what the surrounding ecosystem reads.
-
-**The header is flat.** safetensors metadata is ``str -> str``, so the nested
-provenance block is written one entry per top-level field, with non-strings
-JSON-encoded. Deliberately not a single blob: per-field survives inspection by
-anything that can open the file, which is most of the value. The ONNX sink made
-the same call for the same reason and now shares this encoder.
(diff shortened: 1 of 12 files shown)


YeapJieShen (commented):
Human reviewed, all good.

Issue #243, opened 2026-09-01
SRGAN config skips its parent __post_init__, disabling the compile_mode guard


YeapJieShen:
## The defect

`SRGANTrainingConfig.__post_init__` overrides its parent's and never calls `super()`:

```python
def __post_init__(self) -> None:
    """Reject settings this training mode cannot honour."""
    if self.d_steps_per_g_step < 1:
        raise ValueError(...)
```

`SRTrainingConfig.__post_init__` holds the `compile_mode` / `compile_backend` guard. So for
**every SRGAN config**, that guard does not run.

## Measured

```
SRGANTrainingConfig(compile_mode='max-autotune', compile_backend=None)
-> no raise; compile_mode == 'max-autotune'
```

The same construction on `SRTrainingConfig` or `SRResNetTrainingConfig` raises, as intended.

## What it costs

The guard was added precisely because both failure modes are bad:

- `compile_mode` with **no** backend is dead config — nothing is compiled, and a run that
  looks configured for `max-autotune` trains eager with nothing saying so.
- `compile_mode` with a **non-inductor** backend is passed through to a compiler function
  that does not accept it, so it fails on the **first compiled call** rather than at startup.

SRGAN is the longest-running configuration in the project, so "fails on the first compiled
call" is the most expensive place to lose the startup check.

## Scope

One missing `super().__post_init__()`. An audit of every `__post_init__` in the package finds
this is the **only** override that skips its parent — a one-off, not a pattern.

## Acceptance

- [ ] `SRGANTrainingConfig.__post_init__` calls `super().__post_init__()`
- [ ] The inherited guard fires for SRGAN configs, proven RED first
- [ ] `d_steps_per_g_step` validation is unchanged

Issue #238, opened 2026-09-01
Config fields accept nonsense silently: a negative scale becomes a negative crop border


YeapJieShen:
## The pattern

A separate issue covers `crop_border` accepting a negative value and silently scoring uncropped.
A package-wide sweep — every dataclass field, with inheritance and dynamic `getattr` validation
loops resolved — shows it is **one instance of a pattern**, not a one-off. Each of these is
measured, not inferred:

| field | accepted | consequence |
|---|---|---|
| `training_config.scale = -3` | yes | **resolves `crop_border` to `-3`** on any architecture with no `scale` hparam to disagree with (SRCNN), so the border check never fires and the score is silently uncropped |
| `training_config.scale = 0` | yes | same route, `crop_border = 0` |
| `layer_lrs = [-1e-4, -1e-4, -1e-4]` | yes | every layer trains in the wrong direction; length is checked at optimizer build, values never |
| `init_std = -0.001` | yes | reaches `torch.nn.init.normal_` |
| `adversarial_weight = -1e-3` | yes | inverts the generator's adversarial objective |
| `init_strategy = 'Paper'` | yes | **the paper init silently does not run.** The dispatch is `== "paper"`, so any other string means "default" — no error, no warning |

## Why the last one is the worst

This is a paper-reproduction framework. `init_strategy='Paper'` produces a run that looks
configured for the paper's initialisation and is not, and nothing anywhere says so. The
`Literal` annotation is enforced by jsonargparse at the CLI but by nothing in Python, so a
directly-constructed module — every test, every script — has no guard at all.

## Why the first one matters

It chains. `scale` is not validated, and since the evaluation border derives from `scale` for any
model that does not declare one, a negative `scale` becomes a negative `crop_border`, which is
silently no crop. Two unvalidated fields compose into a wrong published number.

`validate_against` only compares `scale` against the model's own `scale` hparam **when both
exist**. SRCNN declares none — deliberately, it is resolution-preserving — so for SRCNN nothing
checks `scale` at all.

## What is deliberately not in scope

- `separate_psnr` — a bool, nothing to validate.
- `example_input_shape` — already validated in `validate_against` against the real train patch.
- `init_from` — already validated in `setup`, where the file is actually read.

Those three came up in the sweep and are fine.

## Proposal

Validate in `__post_init__` alongside the checks already there, in the same style — a message
naming the YAML key to fix:

- `scale`, when set, must be a positive int
[...]

PR #244, opened 2026-09-01
fix: reject config values that silently produce an incomparable number

### LINKED ISSUE #233: crop_border is unvalidated: a negative value silently scores uncropped
## The gap

`SREvalConfig.__post_init__` validates every field it carries — `psnr_channels`, `ssim_channels`,
`ssim_impl`, `perceptual_metrics`, `lpips_net` — each with a message naming the YAML key to fix.

`crop_border` is validated nowhere.

## What a wrong value does

Measured on two genuinely different 32x32 images:

| `crop_border` | PSNR-RGB | behaviour |
|---:|---:|---|
| `0` | 7.852217 dB | no crop |
| `4` | 7.984293 dB | crops 4 px per side |
| `-4` | **7.852217 dB** | **silently identical to 0 — no crop, no error** |
| `-1` | **7.852217 dB** | same |
| `99` on an 8x8 image | — | raw `RuntimeError` from inside torch, naming no config key |

A negative border produces **a number that looks valid and is not comparable to any published
result.** Two figures here are comparable only if computed the same way, and this one quietly
is not.

## Why it is easy to hit

`None` is now the "derive it from the model's scale" sentinel. So the field's live values are:
`None` (auto), positive (crop), `0` (no crop) — and `-1`, the value someone reaches for when they
mean "auto" and have not read the docstring, sits silently in the same space.

The oversized case is the milder half but the same shape: a config error surfacing as a torch
tensor-dimension `RuntimeError` from several frames down, rather than as the actionable message
every sibling field gets.

## Proposal

Validate in `__post_init__`, matching the existing messages' style. Two rules:

- reject negative values outright — there is no reading under which one is meaningful
- the oversized case cannot be caught at config time (no image is in scope yet), so it belongs at
  scoring time where the tensor is: raise naming `crop_border`, the border, and the image size

## Acceptance

- [ ] A negative `crop_border` raises at construction, naming the YAML key
- [ ] `None`, `0` and positive values are unaffected
- [ ] A border that would consume the whole image raises at scoring time, naming the value and the
      image size, instead of a bare torch `RuntimeError`
- [ ] Tests for all three, proven RED first

YeapJieShen:
**Stack position 1 of 12 · review group: metric-path P2**

Closes #233, #238, #243.

### Why these three are one PR

They are one chain, not three fixes. `crop_border` accepting a negative value is the
defect; `scale` accepting a negative value is *how it gets there*; and #243 is the same
shape one class over — a validation that exists and does not run.

```
training_config.scale = -3         (unvalidated)
  -> SRCNN declares no `scale` hparam, so validate_against compares nothing
  -> SRLightning resolves eval_config.crop_border = -3
  -> SRScorer.crop: `if n <= 0: return tensors`
  -> every PSNR/SSIM scored on the FULL image, no error, no warning
```

That last number looks valid and is comparable to nothing.

### What changed

| field | before | now |
|---|---|---|
| `eval_config.crop_border` | `-4` scored uncropped, identical to `0` | raises, naming the YAML key |
| `training_config.scale` | `-3` / `0` accepted | must be a positive int when set |
| `layer_lrs` | length checked at optimizer build, values never | every entry, and both halves of a pair, must be positive |
| `init_std` | `-0.001` reached `torch.nn.init.normal_` | must be positive (`init_mean` stays free) |
| `init_strategy` | `'Paper'` silently selected PyTorch's default init | must be `'default'` or `'paper'` |
| `adversarial_weight` | `-1e-3` inverted the generator's objective | must be `>= 0` (zero is the content-only ablation) |
| `SRGANTrainingConfig.__post_init__` | replaced its parent's; the `compile_mode` guard never ran | calls `super()` first |

### Two guards, deliberately

`SRLightning` **assigns** `eval_config.crop_border` when it resolves the derive-from-scale
sentinel, which bypasses `__post_init__` entirely. So config validation cannot be the only
guard — `SRScorer.crop` refuses a negative border as well, and the oversized case (which
config time cannot see, because no image is in scope there) raises where the tensor is,
naming both the border and the image size instead of a bare torch `RuntimeError`.

### Evidence

- **9 tests, each proven RED on this branch's parent** before the fix.
- `683 passed` across `tests/training`, `tests/metrics`, `tests/models`.
- **All 34 config objects** across the shipped templates and the run-config mirror still
  construct unchanged — no shipped config changes behaviour.

### On #243

Not from the review passes — found while reading `__post_init__` for this PR. Measured:
`SRGANTrainingConfig(compile_mode='max-autotune', compile_backend=None)` did not raise,
while the identical construction on the parent did. An audit of every `__post_init__` in
the package found this the **only** override skipping its parent.

### DIFF
--- a/sisr/metrics/scoring.py
+++ b/sisr/metrics/scoring.py
@@ -134,6 +134,13 @@ class SRScorer:
                 config never passed through ``SRLightning``, which resolves the
                 derive-from-scale sentinel; scoring against an unresolved border
                 would silently pick a different region than the run intends.
+                Also if it is negative -- reachable only past
+                ``SREvalConfig.__post_init__``, via the sentinel resolution's
+                direct assignment -- or if twice the border would consume an
+                image's shorter axis. The oversized case cannot be caught at
+                config time because no image is in scope there, so it is caught
+                here, where the tensor is, and names both the border and the
+                size rather than surfacing as a torch shape error.
         """
         n = self.eval_config.crop_border
         if n is None:
@@ -142,8 +149,24 @@ class SRScorer:
                 "meaning 'crop the model's scale', resolved by SRLightning at construction "
                 "-- build the scorer from a module's eval_config, or set an explicit int."
             )
-        if n <= 0:
+        if n < 0:
+            raise ValueError(
+                f"eval_config.crop_border is {n} at scoring time. SREvalConfig rejects a "
+                "negative border at construction, but SRLightning ASSIGNS this field when it "
+                "resolves the derive-from-scale sentinel, which bypasses that check -- so a "
+                "non-positive `scale` lands here. Scoring would silently return the FULL "
+                "image, because 'no crop' and 'a negative crop' are the same branch."
+            )
+        if n == 0:
             return tensors
+        for t in tensors:
+            h, w = t.shape[-2:]
+            if 2 * n >= min(h, w):
+                raise ValueError(
+                    f"eval_config.crop_border={n} removes {2 * n} px from each axis of a "
+                    f"{h}x{w} image, leaving nothing to score. Lower crop_border, or drop "
+                    "the images smaller than it from the evaluation set."
+                )
         return tuple(t[..., n:-n, n:-n] for t in tensors)
 
     def metric_tensors(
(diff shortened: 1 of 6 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #246, opened 2026-09-01
fix: modcrop SRCNN's validation HR, matching the authors' degradation order

### LINKED ISSUE #242: SRCNN validation skips the modcrop that its own training grid and SRResNet both apply
## The defect

`sisr/datasets/srcnn.py::ValidationDataset.__getitem__` degrades the **full, uncropped** image:

```python
arr = self._load_rgb(path)
lr_arr = _degrade(arr, self.scale)     # no modcrop
```

`_degrade` downsamples by `scale` with integer division and resizes back:

```python
down = resize(hr_arr, (h // scale, w // scale))
return resize(down, (h, w))
```

When `h % scale != 0` the round trip is not the standard degradation: the image is downsampled to
`h // scale` rows and stretched back over `h` rows, so the LR grid does not align with the HR one.

## The codebase disagrees with itself, three ways

| site | crops to a multiple of `scale` first? |
|---|---|
| `datasets/srcnn.py::_grid_dims` — SRCNN **training** | **yes** — `h_crop = height - (height % scale)` |
| `datasets/srresnet.py::ValidationDataset` | **yes** — `h_crop = h - (h % scale)`, commented *"deterministic exact-corner crop"* |
| `datasets/srcnn.py::ValidationDataset` | **no** |

So SRCNN is degraded one way at train time and another at eval time, and SRResNet is evaluated
differently from SRCNN on the same images.

## Measured

A 481x321 image at scale 3 — BSD100's exact size, where `h % 3 == 1`:

```
ours    -> downsampled to 160x107, stretched back to 481x321
modcrop -> 480x321
max |diff| over the shared region: 51 levels     mean: 2.866
```

**51 levels out of 255.** That is a different input image, not a rounding difference.

## Why it reaches a published number

This is the dataset behind SRCNN's benchmark PSNR/SSIM. At x3, most standard benchmark images are
not divisible by 3 — 481x321 (BSD100) and 512x512 (Set5) both fail — so the affected case is the
common one, not the corner.

The project's standing rule is that a metric figure is comparable only to one computed the same
way. An SRCNN number produced from a non-standard degradation is not comparable to the paper's,
and nothing currently says so.

Nothing is published yet: SRCNN has no publishable row. That makes this cheap to fix now and
expensive to discover after a number ships.

## Scope

Crop to a multiple of `scale` before degrading, matching what this repo already does in the other
two places. The natural home is `_degrade` itself, which is shared by both SRCNN datasets — then
the training grid's own crop becomes redundant rather than load-bearing, and the two cannot drift
again.

**Confirm against the authors' released `demo_SR.m` as part of the fix** rather than assuming: the
`modcrop`-before-degrade convention is standard and both this repo's other sites follow it, but
the authors' file is the primary source and it is already in hand from the earlier evaluation-
geometry work.

## Acceptance

- [ ] SRCNN's validation/test LR is derived from an HR cropped to a multiple of `scale`
- [ ] SRCNN train and eval degrade identically — one implementation, not two conventions
- [ ] A test on a deliberately non-divisible size (481x321 at x3), proven RED first
- [ ] The authors' demo is checked and the finding recorded either way

### LINKED ISSUE #245: SRCNN degrades an extracted patch; the authors degrade the image and then extract
## The defect

SRCNN's training path degrades an **extracted HR patch**. The authors degrade the **whole
image** and extract patches from the result.

Ours, in `datasets/srcnn.py::TrainDataset.__getitem__`:

```python
hr_subimg  = <33x33 slice of the HR image>
lr_subimg  = _degrade(hr_subimg, self.scale)     # degrade the PATCH
```

Theirs, in the released `demo_SR.m` / `generate_train.m` order:

```matlab
im_gnd = modcrop(im, up_scale);
im_l   = imresize(im_gnd, 1/up_scale, 'bicubic');   % degrade the IMAGE
im_b   = imresize(im_l,   up_scale,   'bicubic');
% ... only then are 33x33 sub-images taken from im_gnd / im_b
```

These are not the same operation. Bicubic interpolation's kernel support extends past the
patch boundary, so degrading a 33x33 crop resolves those taps against the crop's edge
handling instead of against the real neighbouring pixels.

## Measured

Smooth, image-like content (noise exaggerates this heavily — on random input the same
comparison reads max 36–59 levels, which is not representative):

| scale | whole patch max\|diff\| | mean | 4 px border mean | interior mean |
|---:|---:|---:|---:|---:|
| 2 | 3 | 0.519 | 0.663 | 0.424 |
| **3** | **5** | 0.333 | **0.831** | **0.000** |
| 4 | 8 | 0.904 | 1.574 | 0.450 |

**At scale 3 the interior is bit-identical and the entire difference is the border.** That is
the signature of the kernel reaching outside the patch and nothing else. At scales 2 and 4 the
interior differs too, because `sub_img_size = 33` is not a multiple of those scales, so the LR
sampling grid also shifts.

A 4 px border is **43%** of a 33x33 patch's pixels.

## Why it is worth filing rather than shrugging at

The differences are small in absolute terms, and this is deliberately filed as a fidelity gap
rather than a correctness bug. But this is a paper-reproduction framework, and the standing
rule is that a figure is comparable only to one computed the same way. Here the *training
inputs* are not produced the way the authors produced theirs — so a reproduction gap has a
candidate explanation that is not the model.

It also means `sub_img_size` and `scale` are silently coupled: `33 % 3 == 0` makes scale 3
much better behaved than 2 or 4, and nothing anywhere says so.

## Relationship to the modcrop issue

A separate issue covers SRCNN *validation* skipping modcrop. That one is about **which pixels
exist**; this one is about **when the degradation is applied relative to patching**. Fixing
the first does not fix this — after it, train and eval agree on the modcrop convention and
still disagree on this axis.

## Scope note

Not free. Degrading whole images at load time, or caching an LR plane alongside the HR one,
changes the data path this project measured at 98.94% decode-dominated — so the fix needs a
throughput measurement, not just a correctness one.

## Acceptance

- [ ] SRCNN training patches are cut from a degraded whole image, matching the released order
- [ ] A test on a patch whose border demonstrably differs today, proven RED first
- [ ] `sub_img_size` / `scale` coupling is either removed or documented where it is chosen

YeapJieShen:
**Stack position 2 of 12 · review group: metric-path P2 · base: `fix/config-field-validation` (#244)**

Closes #242.

### Primary source: checked, not assumed

The issue's acceptance required confirming against the authors' released demo rather than
trusting the convention. Done — and it says exactly what was hoped:

```matlab
im_gnd = modcrop(im, up_scale);
im_l   = imresize(im_gnd, 1/up_scale, 'bicubic');
im_b   = imresize(im_l,   up_scale,   'bicubic');
```

`modcrop` is applied to the **ground truth**, **before** the round trip. So the reference the
metric is computed against is the cropped image, and the HR this dataset serves had to change
too — not just the LR.

*(Correction to the issue: it said the demo was "already in hand from the earlier
evaluation-geometry work". It was not on disk; it was fetched from the public release for this.)*

### Measured

481x321 at x3 — BSD100's exact size, where `481 % 3 == 1`:

```
ours     -> downsampled to 160 cols, stretched back over 481
modcrop  -> 480 cols
max |diff| = 59 / 255      mean = 2.87
```

### One deliberate deviation from the issue's proposal

The issue proposed putting the modcrop **inside `_degrade`**, reasoning that the training
grid's own crop would then become redundant. **I did not do that, and the measurement is why.**

`_degrade` is also called **per sub-image** in the training path. Modcropping there would crop
the *patch*, shrinking a 33x33 sub-image to 32x32 at scales 2 and 4 — silently changing the
training patch size. The authors modcrop the **image**, never the patch.

So the convention gets a single owner, `_modcrop_extent`, which both `_grid_dims` (training)
and `ValidationDataset` (evaluation) derive from. Same guarantee the issue wanted — the two
cannot drift again — without changing the training grid.

### A second, deeper divergence, filed not fixed: #245

While confirming the above I measured something the issue did not cover. We degrade an
*extracted patch*; the authors degrade the *image* and then extract. Bicubic support reaches
past a patch boundary, so these differ:

| scale | max\|diff\| | 4 px border mean | interior mean |
|---:|---:|---:|---:|
| 2 | 3 | 0.663 | 0.424 |
| **3** | **5** | **0.831** | **0.000** |
| 4 | 8 | 1.574 | 0.450 |

At x3 the interior is bit-identical and the whole difference is the border — the signature of
the kernel reaching outside the patch. Filed as **#245** rather than fixed here: it changes
SRCNN's *training* data and the data path is 98.94% decode-dominated, so it needs a throughput
measurement, not just a correctness one.

**This means "train and eval degrade identically" now holds for the modcrop convention and
still does not hold on that second axis.** #245 is the remainder.

### Evidence

- 3 tests proven RED before the fix; 1 more green on both sides, guarding that the
  already-divisible case does not move.
- `474 passed` across `tests/datasets` and `tests/training`.
- Nothing published moves: there is no SRCNN run in this project yet.

### DIFF
--- a/sisr/datasets/srcnn.py
+++ b/sisr/datasets/srcnn.py
@@ -31,6 +31,38 @@ from .base import HRCachedTrainDataset, SRDataset
 from .hr_cache import process_hr_image as _process_hr_image
 
 
+def _modcrop_extent(length: int, scale: int) -> int:
+    """Returns the largest multiple of *scale* not exceeding *length*.
+
+    The single owner of the authors' ``modcrop`` convention, in the one form
+    both callers need: :func:`_grid_dims` wants the cropped extent as a number,
+    :func:`_modcrop` wants to slice by it. Before this existed the convention
+    was spelled out at two sites and absent from a third, which is how SRCNN
+    came to be degraded one way at training time and another at evaluation.
+    """
+    return length - (length % scale)
+
+
+def _modcrop(arr: np.ndarray, scale: int) -> np.ndarray:
+    """Crops *arr* so both spatial axes are whole multiples of *scale*.
+
+    The authors' released ``demo_SR.m`` applies this to the ground truth
+    **before** the bicubic round trip::
+
+        im_gnd = modcrop(im, up_scale);
+        im_l   = imresize(im_gnd, 1/up_scale, 'bicubic');
+        im_b   = imresize(im_l,   up_scale,   'bicubic');
+
+    Without it, an image whose height is not a multiple of *scale* is
+    downsampled to ``h // scale`` rows and stretched back over ``h``, so the LR
+    sampling grid does not align with the HR one. At x3 most standard benchmark
+    images are affected -- 481x321 and 512x512 both are -- so this is the
+    common case, not the corner.
+    """
+    h, w = arr.shape[:2]
+    return arr[: _modcrop_extent(h, scale), : _modcrop_extent(w, scale)]
+
+
 def _grid_dims(
     height: int, width: int, scale: int, sub_img_size: int, stride: int
 ) -> tuple[int, int]:
@@ -39,9 +71,12 @@ def _grid_dims(
     Single source of truth for the deterministic patch grid: :func:`_iter_patch_origins`
     and :meth:`TrainDataset.__getitem__`'s O(1) index -> origin lookup both derive
     positions from this, so they can never silently disagree and misalign an index.
+
+    The extent comes from :func:`_modcrop_extent`, the same function the
+    validation loader crops by, so the two cannot drift apart again.
     """
-    h_crop = height - (height % scale)
-    w_crop = width - (width % scale)
+    h_crop = _modcrop_extent(height, scale)
+    w_crop = _modcrop_extent(width, scale)
     n_rows = len(range(0, h_crop - sub_img_size + 1, stride))
     n_cols = len(range(0, w_crop - sub_img_size + 1, stride))
     return n_rows, n_cols
@@ -209,6 +244,12 @@ class ValidationDataset(SRDataset):
     produced by bicubic-downsampling and bicubic-upsampling back to the
     original size.
 
+    **The HR image is modcropped first** (:func:`_modcrop`), matching the
+    authors' released order and what :class:`TrainDataset`'s grid already did.
+    So the HR served here is up to ``scale - 1`` pixels smaller per axis than
+    the file on disk -- that is the reference the paper's numbers are computed
+    against.
+
     Args:
         img_dir (str | Path): Directory containing the high-resolution
             images.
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #248, opened 2026-09-01
refactor: route every metric tag through the grammar's declared owner

### LINKED ISSUE #230: The metric-tag grammar names a single owner and is bypassed at three sites
## The defect

`metric_tag` in `sisr/metrics/scoring.py` documents itself as the single owner of the metric-tag
grammar:

> The single implementation of the tag grammar. `psnr/val/Y` and `psnr/Set5/Y` differ only in
> `scope`, so routing both through here is what stops the validation and benchmark hierarchies
> drifting apart.

It is called at **two** sites. The same grammar is rebuilt with raw f-strings at **three**:

| site | builds | already available as |
|---|---|---|
| `sisr/training/lightning_module.py:796-798` — `on_train_start`, HParams seeding | `psnr/val/{k}`, `ssim/val/{k}`, `{name}/val` | the key set of `Scores.tagged("val")` |
| `sisr/training/callbacks.py:453,461,469` — `_flush_buffer`, benchmark means | `psnr/{set}/{key}`, `ssim/{set}/{key}`, `{name}/{set}` | `Scores.tagged(dataset_name)` — reproduces this exactly, perceptual special case included |
| `sisr/training/callbacks.py:917-923` — `_validate_monitor` | the whole valid-monitor namespace | `metric_tag` / `Scores.tagged` |

## Why the third one matters most

`_validate_monitor` exists to refuse a checkpoint monitor that names a metric nothing will log —
the failure it prevents is a run that silently keeps its *worst* checkpoint. It answers "which
tags will exist?" by re-deriving the grammar rather than asking the function that owns it.

## How it fails

Change the grammar in `metric_tag`, which is the one place its own docstring says to change it:

- `validation_step` emits the new tags; `_validate_monitor` still validates against the old ones,
  so a correct monitor is rejected or an incorrect one accepted.
- `_flush_buffer` keeps emitting the old shape, so the benchmark hierarchy and the validation
  hierarchy diverge — exactly the drift `metric_tag`'s docstring says routing through it prevents.

**Nothing in the suite would catch either.** No test asserts that the tags `validation_step`
emits are the tags `_validate_monitor` accepts.

## Scope

Three call sites. No interface change, no behaviour change when the grammar is unchanged — which
is the point: the change is invisible today and load-bearing the moment the grammar moves.

`_flush_buffer` is the clearest: it hand-rolls what `Scores.tagged(dataset_name)` already returns,
including the perceptual tags' different shape.

## Acceptance

- [ ] The three sites build tags through `metric_tag` / `Scores.tagged`
- [ ] A test that fails if the emitted tag set and the validator's accepted set disagree — this is
      the part that keeps it fixed, and it is missing today. Prove it RED by changing the grammar
      in one place.
- [ ] `per_image/...` tags are out of scope; they are a genuinely different, deeper hierarchy

YeapJieShen:
**Stack position 3 of 12 · review group: metric-path P2 · base: `fix/srcnn-validation-modcrop` (#246)**

Closes #230.

### The grammar has two rules, not one

The issue said "route the three sites through `metric_tag`". That is not sufficient on its
own, and the reason is the bug's own shape: perceptual metrics tag as `{name}/{scope}`, with
no colorspace key, because LPIPS/DISTS have no colorspace decomposition. **A caller
re-deriving "the tag grammar" from `metric_tag` alone silently gets that family wrong — which
is exactly what `_flush_buffer` did.**

So this adds a second owner, `perceptual_tag`, and `expected_tags` built from both:

| function | answers |
|---|---|
| `metric_tag` | `{family}/{scope}/{key}` — one tag |
| `perceptual_tag` | `{name}/{scope}` — one tag, different shape, already published |
| `Scores.tagged` | which tags exist, **with** values |
| `expected_tags` | which tags exist, **without** values |

That last one is what the three bypass sites needed and did not have.

### The three sites

| site | was | now |
|---|---|---|
| `on_train_start` HParams seeding | three f-string dict comprehensions | `expected_tags(eval_config, "val")` |
| `_flush_buffer` benchmark means | hand-rolled per family, perceptual special-cased | mean each family, then `Scores.tagged(dataset_name)` |
| `_validate_monitor` | rebuilt the whole namespace | `expected_tags` + `perceptual_tag` for the direction split |

### The test that keeps it fixed

The issue asked for "a test that fails if the emitted tag set and the validator's accepted set
disagree — prove it RED by changing the grammar in one place." That is what the three tests do:
they **move the grammar in its owner** and assert the other sites follow.

Proven RED. Before the fix, with the grammar moved:

```
benchmark tags did not follow the grammar:
  ['psnr/Set5/RGB', 'psnr/Set5/Y', 'ssim/Set5/RGB', 'ssim/Set5/Y']
  expected 'psnr.Set5.Y'
```

### One type widening, deliberately

`Scores` values become `Tensor | float`. The validation path carries 0-dim tensors straight
from the metric functions; the benchmark path carries means it computed itself. `Scores` is a
carrier and never does arithmetic on them, so it accepts both rather than forcing one side to
convert for the type checker. The single site that assumed tensors now uses `float()` — exact
for a 0-dim tensor, total over both.

### Evidence

- 3 tests proven RED before the change.
- `599 passed` across `tests/training` and `tests/metrics`.
- `mypy`: clean over 48 source files. `ruff check` + `ruff format --check`: clean.
- **No behaviour change while the grammar is unchanged** — which is the point. It is invisible
  today and load-bearing the moment the grammar moves.

### DIFF
--- a/sisr/metrics/scoring.py
+++ b/sisr/metrics/scoring.py
@@ -63,6 +63,64 @@ def metric_tag(family: str, scope: str, key: str) -> str:
     return f"{family}/{scope}/{key}"
 
 
+#: A metric value as it travels to ``pl_module.log``. The validation path
+#: produces 0-dim tensors straight from the metric functions; the benchmark
+#: path means its per-image buffer first and so produces plain floats. Both are
+#: legal ``log`` arguments, and ``Scores`` is only a carrier -- it never does
+#: arithmetic on them -- so it accepts either rather than forcing one side to
+#: convert for the type checker's benefit.
+ScoreValue = torch.Tensor | float
+
+
+def perceptual_tag(name: str, scope: str) -> str:
+    """Build a perceptual metric's tag: ``{name}/{scope}``.
+
+    The second rule of the grammar, and deliberately a different shape:
+    LPIPS/DISTS carry no colorspace key, so they get their own family root
+    rather than a key slot under the PSNR/SSIM scheme. That shape is already
+    published, so it is preserved here rather than normalised -- which is
+    exactly why it needs an owner too. A caller re-deriving "the tag grammar"
+    from :func:`metric_tag` alone silently gets this family wrong.
+
+    Args:
+        name: Perceptual metric name, e.g. ``'lpips'``.
+        scope: ``'val'`` for the validation loop, else a benchmark dataset name.
+
+    Returns:
+        The TensorBoard tag.
+    """
+    return f"{name}/{scope}"
+
+
+def expected_tags(eval_config: SREvalConfig, scope: str) -> list[str]:
+    """Every tag a run under *eval_config* will log at *scope*.
+
+    The values-free half of the grammar: :meth:`Scores.tagged` answers "which
+    tags exist" **with** numbers, this answers it without any. Both derive from
+    :func:`metric_tag` and :func:`perceptual_tag`, so a caller that needs to
+    know the tag namespace before a single batch has been scored -- seeding
+    TensorBoard's HParams columns, or validating a checkpoint monitor -- asks
+    the owner instead of re-deriving the grammar from f-strings.
+
+    That re-derivation is not hypothetical: the checkpoint monitor validator,
+    whose whole job is knowing which tags will exist, had its own copy. When
+    the two disagree the failure is silent -- a correct monitor is rejected, or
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

Issue #232, opened 2026-09-01
gan_module imports a private symbol from metadata


YeapJieShen:
## The issue

`sisr/training/gan_module.py` imports a private symbol across a module boundary:

```python
from .metadata import _class_path, build_metadata
```

`_class_path` is private by name and public by use. That combination is the problem: the
underscore says "free to change", the second caller says otherwise, and nothing reconciles them.

## Either direction is fine; the current state is not

- **Promote it.** If a second module legitimately needs "dotted `module.ClassName` for an
  instance", it is part of `metadata`'s interface and should lose the underscore.
- **Remove the need.** If `gan_module` only wants it to build a comparison against
  `build_metadata`'s own output, it may be reachable through that function instead.

## Why bother

It is small. It is filed because a private helper with an outside caller is how a helper quietly
becomes unchangeable — the next person to touch `metadata` has no way to know from the signature
that a second module depends on this one.

## Acceptance

- [ ] `_class_path` is either public, or not imported outside `metadata`
- [ ] No behaviour change

PR #249, opened 2026-09-01
refactor: one scale resolution, and it always returns an int

### LINKED ISSUE #236: Scale resolution is implemented twice and the two already disagree on type
## The duplication

Two implementations of the same three-step resolution — `training_config.scale`, else the model's
`scale` hparam, else raise:

- `SRLightning._resolved_scale`
- `sisr/training/metadata.py::build_metadata`

## They already differ

`_resolved_scale` returns `int(scale)`. `build_metadata` does not coerce. So a config carrying
`scale: "4"` writes a **string** into artifact provenance and an **int** into the evaluation crop
border — the same fact, recorded two ways, from one config.

Artifact provenance is the thing a downstream consumer reads to know what the file means, so a
type that varies with how the YAML was written is not a good property for it to have.

## History

`_resolved_scale` is recent — it arrived with the change that made `crop_border` derive from
`scale`. `build_metadata`'s copy predates it and was not folded in at the time. This is fresh
duplication, not an old one.

## Note on the fix

`metadata` calling `module._resolved_scale(...)` would repeat the private-symbol-across-modules
problem filed separately. `_resolved_scale` should lose its underscore as part of this.

## Acceptance

- [ ] One implementation, called by both
- [ ] The returned scale is an `int` on both paths
- [ ] A test that a string-valued `scale` in config yields an int in artifact metadata, proven RED

YeapJieShen:
**Stack position 4 of 12 · review group: metric-path P2 · base: `refactor/metric-tag-single-owner` (#248)**

Closes #236. Last of the metric-path group.

### The duplication and the divergence

| | `SRLightning._resolved_scale` | `metadata.build_metadata` |
|---|---|---|
| resolution | config → model hparam → raise | *identical* |
| coercion | `int(scale)` | **none** |

One config, one fact, two recorded types: a `str` in artifact provenance and an `int` in the
evaluation crop border.

`build_metadata` now calls the resolver, which **loses its underscore** in the same change —
the issue's own note, and the same reasoning #232 applies one module over.

### Where the string actually comes from

The issue framed this as `scale: "4"` in YAML. Two commits down this stack, `scale` became
type-checked, so that route is now closed at construction. **The remaining route is the model's
own `scale` hparam**, which is free-form and type-checked by nothing — so the test sources the
string from there. Same defect, the route that is still open.

### A defect introduced earlier in this stack, fixed here

The `scale` validation added in #244 compared with `<`:

```
SRTrainingConfig(scale='4')
-> TypeError: '<' not supported between instances of 'str' and 'int'
```

A bare `TypeError` naming no config key — precisely the failure that validation exists to
remove. It now type-checks first, **excluding `bool` explicitly**: `isinstance(True, int)` is
`True`, and `scale: true` in YAML is not an upscaling factor.

Fixed here rather than by rewriting the earlier commit, because scale's *type* is this change's
subject and rewriting would have rebuilt three PRs' worth of stack for it.

### Evidence

- 4 tests proven RED first (3 for #236, 1 for the `TypeError`).
- One of them is structural: it asserts `build_metadata` no longer contains
  `hparams.get("scale")`, so a re-introduced copy fails rather than silently diverging again.
- `871 passed, 2 skipped` (the 2 are Windows-only liveness paths).
- `mypy` clean over 48 files; `ruff check` + `format --check` clean.

### DIFF
--- a/sisr/training/config.py
+++ b/sisr/training/config.py
@@ -145,9 +145,12 @@ class SRTrainingConfig:
                 ``init_std`` is not positive; or if ``init_strategy`` is
                 neither ``'default'`` nor ``'paper'``.
         """
-        if self.scale is not None and self.scale < 1:
+        if self.scale is not None and (
+            not isinstance(self.scale, int) or isinstance(self.scale, bool) or self.scale < 1
+        ):
             raise ValueError(
-                f"training_config.scale must be a positive int; got {self.scale}. "
+                f"training_config.scale must be a positive int; got {self.scale!r} "
+                f"({type(self.scale).__name__}). "
                 "It is not only provenance: for an architecture that declares no `scale` "
                 "hparam of its own (SRCNN, deliberately -- it is resolution-preserving) "
                 "nothing else checks it, and eval_config.crop_border derives from it, so a "
--- a/sisr/training/lightning_module.py
+++ b/sisr/training/lightning_module.py
@@ -100,7 +100,7 @@ class SRLightning(lightning.LightningModule):
         # Resolved here, once, BEFORE the scorer is built -- the scorer and
         # every artifact's sisr_meta must record the number actually used.
         if self.eval_config.crop_border is None:
-            self.eval_config.crop_border = self._resolved_scale(
+            self.eval_config.crop_border = self.resolved_scale(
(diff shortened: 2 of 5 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #250, opened 2026-09-01
fix: an unbound VGG loss refuses to compute instead of assuming [0, 1]

### LINKED ISSUE #239: An unbound VGG loss silently assumes a [0,1] output range
## The defect

`_VGGFeatureLoss.__init__` sets

```python
self._gain, self._offset = 1.0, 0.0
```

and `bind(processor)` later replaces them from `processor.output_range`. So a loss that was never
bound **silently behaves as though the model emits `[0, 1]`**, rather than failing.

Verified:

```
forward WITHOUT bind() -> succeeded, value 0.1684
_gain=1.0 _offset=0.0
```

## Why the default is the wrong choice here

The identical question was already answered the other way, one seam over.
`SRProcessor.output_range` is **abstract rather than defaulted**, and its docstring says why:

> Abstract, never defaulted: a wrong inherited default is the failure this exists to prevent.
> Reconstructing `[-1, 1]`-trained weights with `[0, 1]` logic produces silently wrong images and
> no error.

`SRLoss.bind` is abstract for the same stated reason. Yet the consumer of that carefully-abstract
value gives itself the identity as a fallback — which makes **"never bound"** and **"bound to a
`[0, 1]` processor"** indistinguishable at runtime.

## When it bites

Not through any shipped config: `SRLightning` binds an `SRLoss` criterion, and
`WeightedSumLoss.bind` forwards to its `SRLoss` terms, so both live paths bind.

It bites when a VGG term is reached through a container that is **not** itself an `SRLoss` — a
plain `nn.ModuleList`, or a user's own composite. Paired with `RGBSignedOutputProcessor`, whose
range is `[-1, 1]`, the loss then feeds `[-1, 1]` data through ImageNet normalisation designed for
`[0, 1]`. Training proceeds, the number looks plausible, and the features mean something other
than what the recipe says.

That is the flagship reproduction's content loss.

## Proposal

Make "never bound" loud, matching the seam's own stated reasoning: initialise to a sentinel and
raise in `_features` if it is still there, naming `bind` and the two live paths that call it.

The error should say what to do — put the term inside `WeightedSumLoss`, or call `bind` — because
the person who hits this has written a custom composite and needs to know the contract exists.

## Acceptance

- [ ] An unbound `_VGGFeatureLoss.forward` raises, naming `bind`
- [ ] Every shipped path is unaffected: direct `SRLoss` criterion, and nested in `WeightedSumLoss`
- [ ] A test proving the unbound call raises, RED before the change

YeapJieShen:
**Stack position 5 of 12 · review group: correctness · base: `refactor/one-scale-resolution` (#249)**

Closes #239. First of the correctness group.

### The asymmetry, in one line each

```python
# sisr/processors/base.py -- abstract, never defaulted
output_range: "a wrong inherited default is the failure this exists to prevent"

# sisr/losses/vgg.py -- the consumer of that value, before this PR
self._gain, self._offset = 1.0, 0.0     # the identity
```

`SRProcessor.output_range` is abstract precisely so nobody inherits a wrong range.
`SRLoss.bind` is abstract for the same stated reason. And then the loss handed itself the
identity anyway — making **"never bound"** and **"bound to a `[0, 1]` processor"**
indistinguishable at runtime.

### It does not fire on any shipped path

`SRLightning` binds an `SRLoss` criterion; `WeightedSumLoss.bind` forwards to its `SRLoss`
terms. Both live paths bind, and both have a test here proving they still do.

It bites through a container that is **not** itself an `SRLoss` — a plain `nn.ModuleList`, or a
user's own composite. Paired with `RGBSignedOutputProcessor` (range `[-1, 1]`), the loss feeds
`[-1, 1]` data through ImageNet normalisation designed for `[0, 1]`. Training proceeds, the
number looks plausible, and the features mean something other than the recipe says.

That is the flagship reproduction's content loss.

### The test that stops this being theatre

Four tests cover the mechanism. A fifth pins the **consequence**:

```python
def test_signed_range_processor_actually_changes_the_mapping(vgg22):
    vgg22.bind(RGBProcessor());            as_unit   = vgg22(x, y)
    vgg22.bind(RGBSignedOutputProcessor()); as_signed = vgg22(x, y)
    assert not torch.isclose(as_unit, as_signed)
```

If `[0, 1]` and `[-1, 1]` logic agreed, the silent default would have cost nothing and none of
this would be worth doing. They do not agree.

### Two existing tests edited, and why that is not a smell

`test_feature_scale_squares_into_an_mse_and_is_linear_in_an_l1` and
`test_before_activation_changes_the_features` both called the loss **unbound**, relying on the
identity. Each now binds **identically on both sides** of its comparison, so each still
isolates exactly the one variable it was written to isolate. The contract changed on purpose;
these are the two call sites that were exercising the removed default.

### Evidence

- 2 tests proven RED before the change; 3 more green on both sides as guards.
- `495 passed` across `tests/training` and `tests/losses`.
- `mypy` clean; `ruff check` + `format --check` clean.

### DIFF
--- a/sisr/losses/vgg.py
+++ b/sisr/losses/vgg.py
@@ -171,8 +171,15 @@ class _VGGFeatureLoss(SRLoss):
         self.distance = distance
         self.grayscale_to_rgb = grayscale_to_rgb
         self.allow_non_rgb = allow_non_rgb
-        # Identity until bind(); a [0, 1] processor needs no mapping anyway.
-        self._gain, self._offset = 1.0, 0.0
+        # Sentinel, NOT the identity. An identity default would make "never
+        # bound" and "bound to a [0, 1] processor" indistinguishable at
+        # runtime, and the wrong one of those trains on features that mean
+        # something other than the recipe says. SRProcessor.output_range --
+        # the value bind() reads -- is abstract rather than defaulted for
+        # exactly this reason, and its consumer must not undo that by handing
+        # itself a fallback.
+        self._gain: float | None = None
+        self._offset: float | None = None
         # persistent=False: constants, and a distributable checkpoint should not
         # carry them.
         self.register_buffer(
@@ -188,7 +195,15 @@ class _VGGFeatureLoss(SRLoss):
         return applied
 
     def bind(self, processor: SRProcessor) -> None:
-        """Derive the affine map from the model's output range into ``[0, 1]``."""
+        """Derive the affine map from the model's output range into ``[0, 1]``.
+
+        Until this runs the loss refuses to compute -- see :meth:`_features`.
+
+        Raises:
+            ValueError: If the processor emits a channel count or colorspace
+                this loss has no published recipe for, or a non-increasing
+                ``output_range``.
+        """
         channels = processor.model_channels
         if channels != 3 and not self.grayscale_to_rgb:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #251, opened 2026-09-01
fix: honour SRResNet's padding everywhere, and reject one that cannot work

### LINKED ISSUE #241: SRResNet records a padding hparam that its upsample convs ignore
## The defect

`SRResNet.__init__` takes `padding` and threads it into the head, the residual blocks and the
tail. `SRUpsampleBlock` hardcodes it:

```python
torch.nn.Conv2d(channels, channels * (scale**2), kernel_size=kernel_size, padding="same")
```

Measured on `SRResNet(scale=2, padding=0)`:

| padding in effect | conv count | example |
|---|---:|---|
| `(0, 0)` — as configured | **35** | `head.0` |
| `same` — ignored the argument | **1** | `upsample.0.upsample.0` |

## Why it is worth fixing even though nothing hits it today

Every shipped config uses `padding: same`, so the inconsistency is currently invisible. Two
reasons it still matters:

1. **The recorded hyperparameter becomes a lie.** `padding` goes into `_hparams`, and `_hparams`
   goes verbatim into checkpoint and ONNX provenance as `model.init_args`. A file would claim
   `padding: 0` while one of its convolutions was built `same`. This project treats artifact
   metadata as the thing a downstream consumer trusts to know what a file means.
2. **It is silent.** Nothing raises, nothing warns; the model simply mixes two padding regimes.
   The output shape still works out, so the first sign would be an unexplained metric.

Contrast `SRCNN`, where `padding` reaches every convolution, and where a mode that could not be
expressed (`eval_padding='same'` with an even kernel) raises rather than silently doing something
else.

## The judgement call the fix needs

`SRUpsampleBlock` feeds `PixelShuffle`, which needs its input to be exactly `channels * scale**2`
at the same spatial size — so a valid convolution there shrinks the feature map before the shuffle
and changes the output geometry. It may be that `same` is genuinely required here rather than
merely convenient.

If so, **the fix is to say that, not to thread the argument through**: `SRUpsampleBlock` should
document why it is fixed, and `SRResNet` should either reject a non-`same` `padding` outright or
record what was actually built. What it must not do is accept an argument, record it, and ignore it.

## Acceptance

- [ ] `padding` is either honoured everywhere, or rejected/normalised where it cannot be
- [ ] `_hparams["padding"]` describes what was actually constructed
- [ ] A test asserting every `Conv2d` in the built model agrees with the recorded hparam

YeapJieShen:
**Stack position 6 of 12 · review group: correctness · base: `fix/vgg-loss-requires-bind` (#250)**

Closes #241.

### The judgement call, settled by measurement

The issue flagged that the fix needed a judgement — and guessed `same` might be genuinely
required by `PixelShuffle`. Measured first, on `SRResNet(scale=2)` with the paper's 9-3-9
kernels:

| `padding` | result |
|---|---|
| `'same'` | OK — 24x24 → 48x48 |
| `0` | **`RuntimeError`** — the residual skip adds a 16-wide tensor to a 12-wide one |
| `1` | **"works", returns 30x30 from 24x24 at scale 2** |

So the knob was broken for *every* value but `same`: one raised from several frames inside
`forward`, the other silently produced a model that is no longer `xscale` (head and tail each
shrink by 6 with a 9-kernel and `padding=1`).

### So the accepted set is a rule, not a value

Every convolution must be **shape-preserving**, for two independent reasons: the residual
blocks add `x` to their own output, and `PixelShuffle` multiplies whatever the upsample conv
loses by `scale`. That means `'same'`, **or** `(k - 1) // 2` for every kernel the model builds —
the upsample block's included.

```
padding=1, kernel_sizes=(3,3,3)  ->  legal, and still 24x24 -> 48x48
padding=1, kernel_sizes=(9,3,9)  ->  ValueError naming the kernel it cannot preserve
padding=0, any                   ->  ValueError, not a RuntimeError from inside forward
padding='same'                   ->  unchanged
```

This keeps the knob meaningful rather than reducing it to a one-value field, and it is why
`SRUpsampleBlock` now **takes** `padding` instead of hardcoding it — which is what makes
`_hparams["padding"]` describe what was actually constructed.

### Why the recorded hparam mattered even though nothing hits it

`_hparams` goes verbatim into checkpoint and ONNX provenance as `model.init_args`. A file
could claim `padding: 0` while one of its convolutions was built `same`. This project treats
artifact metadata as the thing a downstream consumer trusts to know what a file means.

### Evidence

- 3 tests proven RED; 2 more green on both sides (the `'same'` path, and the legitimate
  `padding=1` + all-3-kernels case that must not be caught by the new rule).
- The acceptance test walks **every `Conv2d` in the built model** and asserts it agrees with
  the recorded hparam — the upsample block included, which is the one that used to disagree.
- **All 14 shipped SRResNet/SRGAN generator configs** pass the new rule unchanged.
- `507 passed` across `tests/models` and `tests/training`; `mypy`, `ruff` clean.

### DIFF
--- a/sisr/models/srresnet/model.py
+++ b/sisr/models/srresnet/model.py
@@ -50,21 +50,42 @@ class SRResidualBlock(torch.nn.Module):
         return identity + x
 
 
+#: Kernel size of the sub-pixel upsample convolution. Fixed by the architecture
+#: rather than exposed, but named so the padding check can see it.
+_UPSAMPLE_KERNEL = 3
+
+
 class SRUpsampleBlock(torch.nn.Module):
     """Sub-pixel convolution upsampling block: conv -> PixelShuffle -> PReLU.
+
+    The conv must be **shape-preserving**: ``PixelShuffle`` turns
+    ``(C*scale**2, H, W)`` into ``(C, H*scale, W*scale)``, so any spatial size
+    the conv loses is multiplied by ``scale`` in the output. It does not have to
+    be the literal string ``'same'`` -- ``p == (kernel_size - 1) // 2`` is the
+    same map for an odd kernel -- and :class:`SRResNet` enforces exactly that
+    before constructing one. This used to hardcode ``'same'`` and ignore the
+    ``padding`` its parent recorded.
 
     Args:
         channels: Input channel count.
         scale: Upscaling factor. Defaults to ``2``.
         kernel_size: Kernel size for the conv layer. Defaults to ``3``.
+        padding: Padding for the conv layer. Must be shape-preserving.
+            Defaults to ``'same'``.
     """
 
-    def __init__(self, channels: int, scale: int = 2, kernel_size: int = 3):
+    def __init__(
+        self,
+        channels: int,
+        scale: int = 2,
+        kernel_size: int = 3,
+        padding: str | int = "same",
+    ):
         super().__init__()
 
         self.upsample = torch.nn.Sequential(
             torch.nn.Conv2d(
-                channels, channels * (scale**2), kernel_size=kernel_size, padding="same"
+                channels, channels * (scale**2), kernel_size=kernel_size, padding=padding
             ),
             torch.nn.PixelShuffle(scale),
             torch.nn.PReLU(),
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #252, opened 2026-09-01
fix: abort a failed LMDB write transaction instead of leaving it to the GC

### LINKED ISSUE #240: Cleanup calls sit outside with/finally in four places, and one leaks unboundedly
## The asymmetry

`sisr/utils/cache.py` opens six LMDB transactions. Four are context-managed:

```python
with env.begin(write=False, buffers=True) as txn:
```

Two are not — and both are the **write** transactions, where aborting on failure is what matters:

- `LMDBCacheBuildContext.write_batch` — `txn = self.env.begin(write=True)` ... `txn.commit()`
- `LMDBCache._build` — same shape

## `_build` is defensible; `write_batch` is not

`_build`'s transaction sits inside `try: ... finally: env.close()`, and closing the environment
invalidates the open transaction. It is bounded.

`write_batch` has no such bound. Its environment is owned by the caller and outlives it, and
`parallel_build` calls it in a **loop** — once per completed job.

## The failure is anticipated elsewhere in the codebase

`hr_cache.estimate_map_size` exists precisely because of this failure mode, and says so:

> ... `MapFullError` mid-build and the checksum is never written

So a `put` raising `MapFullError` partway through `write_batch` is a known, designed-for event.

## What actually happens, measured

The transaction is aborted **only by garbage collection**, so it survives exactly as long as the
exception's traceback does:

| scenario | next write |
|---|---|
| exception raised, caught, `except` block exits (traceback released) | **succeeds** — refcounting collected the txn |
| a reference to the failed transaction is still held | **blocked**, `lmdb.Error` |

A retained traceback is the second case. Anything that logs with `exc_info=True`, stores the
exception, wraps and re-raises it, or holds it while a `tqdm` bar and a pool of futures are still
live keeps the write lock held — which is exactly the state `parallel_build` is in when a job
fails.

So this is not an unconditional bug, and the issue does not claim one. It is a correctness
property left to the garbage collector in the one place the codebase already expects to fail.

## Also here, smaller

`parallel_build` calls `pbar.close()` only on the success path — both in the inline branch and
after the pool. An exception leaves the `tqdm` bar open, which corrupts subsequent terminal
output. A `try/finally`, or using `tqdm` as a context manager, fixes both branches.

## Acceptance

- [ ] `write_batch`'s transaction is context-managed or `finally`-aborted
- [ ] A test that a failed `write_batch` leaves the environment writable **while a reference to
      the exception is still held** — the weaker version passes today and proves nothing
- [ ] The progress bar is closed on the exception path too

YeapJieShen:
**Stack position 7 of 12 · review group: correctness · base: `fix/srresnet-padding-honoured` (#251)**

Closes #240. Last of the correctness group.

### The asymmetry

`cache.py` opens six LMDB transactions. Four were context-managed. The two that were not were
**both the write transactions** — where aborting on failure is the thing that matters.

`_build`'s is defensible: it sits inside `try: ... finally: env.close()`, and closing the
environment invalidates the transaction. It is bounded. **`write_batch`'s is not** — its
environment is owned by the caller and outlives it, and `parallel_build` calls it in a loop,
once per completed job.

### Why the test's exact shape is the point

The issue was explicit that the weak version of this test proves nothing, and it is right.
Uncontext-managed, the transaction is aborted **only by garbage collection**, so it survives
exactly as long as the exception's traceback:

| scenario | next write |
|---|---|
| `except` block exits, traceback released | **succeeds** — refcounting collected it |
| a reference to the exception is still held | **blocked** |

So the test holds the exception. Proven RED with that shape:

```
lmdb.Error: Attempt to start a write transaction while another write
transaction is active on the same thread. This would deadlock.
```

A retained traceback is not exotic — anything that logs with `exc_info=True`, stores the
exception, wraps and re-raises it, or merely holds it while a `tqdm` bar and a pool of futures
are still live is in that state. **That is exactly the state `parallel_build` is in when a job
fails.** And a `put` raising `MapFullError` partway through is a designed-for event here:
`hr_cache.estimate_map_size` exists because of it and says so.

*(The test's map sizing was tuned after the first attempt: at 16 KB the post-abort write failed
for genuine exhaustion rather than for the lock, which would have made it pass for the wrong
reason. It was re-proven RED at the final sizing.)*

### Also here

`parallel_build` called `pbar.close()` only on the success path, in **both** branches, so an
exception left the bar open and corrupted every subsequent line of terminal output. Now a
`try/finally` covering both.

### Evidence

- 2 tests, both proven RED, and the first re-proven RED after its sizing changed.
- `138 passed, 2 skipped` across `tests/utils` and `tests/datasets` (the 2 are Windows-only).
- `mypy` clean; `ruff` clean.

### DIFF
--- a/sisr/utils/cache.py
+++ b/sisr/utils/cache.py
@@ -129,13 +129,26 @@ class LMDBCacheBuildContext:
     def write_batch(self, pairs: Sequence[tuple[str, bytes]]) -> None:
         """Writes a batch of key-value pairs in a single transaction.
 
+        Context-managed, so a failed put aborts the transaction immediately.
+        Uncontext-managed it was aborted only by garbage collection, which
+        means it survived exactly as long as the raised exception's traceback
+        -- and anything that logs with ``exc_info=True``, stores the exception,
+        wraps and re-raises it, or merely holds it while a ``tqdm`` bar and a
+        pool of futures are still live keeps the write lock held. That is
+        precisely the state :meth:`parallel_build` is in when a job fails.
+
+        The environment here is owned by the caller and outlives this method,
+        and ``parallel_build`` calls it in a loop, so nothing else bounds it.
+        A ``put`` raising ``MapFullError`` partway through is a known,
+        designed-for event -- ``hr_cache.estimate_map_size`` exists because of
+        it.
+
         Args:
             pairs: Sequence of ``(key, value)`` tuples to write.
         """
-        txn = self.env.begin(write=True)
-        for key, value in pairs:
-            txn.put(key.encode(), value)
-        txn.commit()
+        with self.env.begin(write=True) as txn:
+            for key, value in pairs:
+                txn.put(key.encode(), value)
 
     def parallel_build(
         self,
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #253, opened 2026-09-01
refactor: give _collect_batch's two guarded emissions their own methods

### LINKED ISSUE #231: BenchmarkImageLogger._collect_batch does four things in one 82-line loop
## What is oversized

`BenchmarkImageLogger._collect_batch` — **82 code lines**, one loop, four responsibilities behind
three different gates:

| responsibility | gate |
|---|---|
| score each image through the shared scorer, buffer the result | always |
| emit `per_image/...` PSNR/SSIM scalars | `log_per_image_metrics` |
| move three tensors to host, build and emit the bicubic-SR-HR triptych | `should_log_images` |
| (all of the above except buffering) | `self._tb_experiment is not None` |

The tell is already in the file: the triptych block carries a **nine-line comment explaining what
it does**, which is a comment describing a function that has not been written yet.

## What this does not propose

`callbacks.py` stays one file. The module groups callbacks by role and splitting it would trade
one long file for several short ones plus an import graph. **The oversized unit is a method, not
the module.**

## Proposal

Two extractions from the per-image loop, no interface change and no behaviour change:

- `_emit_per_image_scalars(dataset_name, filename, psnr, ssim, step)`
- `_emit_triptych(dataset_name, filename, lr_cpu, sr_cpu, hr_cpu, step)`

`_collect_batch` drops to roughly 35 lines and each gate becomes legible where it is taken. The
nine-line comment becomes the extracted function's docstring, which is where it belongs.

## Acceptance

- [ ] `_collect_batch` reads as: score, buffer, then two guarded emissions
- [ ] Existing benchmark-logging tests pass unchanged — this is a refactor, so the evidence is
      that no test needed editing
- [ ] The bicubic/pad behaviour is untouched: SR is centre-padded only when smaller than HR, and
      the bicubic panel still resamples at 1:1 for pre-upsampled architectures

YeapJieShen:
**Stack position 8 of 12 · review group: hygiene / UX · base: `fix/lmdb-write-txn-abort` (#252)**

Closes #231. First of the hygiene group.

### What was oversized

`_collect_batch` did four things in one loop behind three different gates:

| responsibility | gate |
|---|---|
| score through the shared scorer, buffer the result | always |
| emit `per_image/...` PSNR/SSIM scalars | `log_per_image_metrics` |
| host-transfer three tensors, compose and emit the triptych | `should_log_images` |
| (all of the above except buffering) | `self._tb_experiment is not None` |

**The tell was already in the file**: the triptych block carried a nine-line comment explaining
what it does — a comment describing a function that had not been written yet. It is now that
function's docstring.

### Measured

```
_collect_batch body:  74 -> 52 lines   (22 moved out, 29% smaller)
  _emit_per_image_scalars   7 lines
  _emit_triptych            8 lines
```

Each gate is now legible where it is taken, rather than several screens from what it guards.

### What this does not do

`callbacks.py` stays **one file**. The module groups callbacks by role and splitting it would
trade one long file for several short ones plus an import graph. The oversized unit was a
method.

### Evidence

**No test needed editing** — for a refactor that is the evidence, and it is the acceptance
criterion the issue named. `git diff --name-only -- tests/` is empty.

- `120 passed` in `tests/training/test_callbacks.py`, unchanged.
- That set includes the tests pinning the two behaviours the issue asked to leave untouched:
  SR is centre-padded **only when smaller than HR**, and the bicubic panel resamples at 1:1 for
  pre-upsampled architectures.
- `mypy` clean; `ruff check` + `format --check` clean.

### DIFF
--- a/sisr/training/callbacks.py
+++ b/sisr/training/callbacks.py
@@ -401,37 +401,80 @@ class BenchmarkImageLogger(Callback):
                 continue
 
             if self.log_per_image_metrics:
-                for key, psnr_val in psnr_dict.items():
-                    tag = f"per_image/{dataset_name}/psnr/{key}/{filename}"
-                    self._tb_experiment.add_scalar(tag, psnr_val, global_step=step)
-                for key, ssim_val in ssim_dict.items():
-                    tag = f"per_image/{dataset_name}/ssim/{key}/{filename}"
-                    self._tb_experiment.add_scalar(tag, ssim_val, global_step=step)
+                self._emit_per_image_scalars(dataset_name, filename, psnr_dict, ssim_dict, step)
 
             if not should_log_images:
                 continue
 
             # One host transfer per image, only on cycles that log strips.
-            lr_cpu = lr_img[i].cpu()
-            sr_cpu = sr[i].cpu()
-            hr_cpu = hr_img[i].cpu()
-
-            # Triptych: bicubic | SR | HR, all at HR size.
-            # The bicubic panel is the LR upsampled to HR via bicubic
-            # interpolation — the standard SR baseline.  For SRCNN the
-            # LR is already at HR size (pre-upsampled in the dataset),
-            # so this resamples at 1:1 and is near-identity.  For
-            # SRResNet (LR < HR) it upscales to HR for direct visual
-            # comparison against SR.  SR is center-padded only when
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #254, opened 2026-09-01
refactor: promote metadata.class_path, which had an outside caller already

### LINKED ISSUE #232: gan_module imports a private symbol from metadata

YeapJieShen:
**Stack position 9 of 12 · review group: hygiene / UX · base: `refactor/collect-batch-extractions` (#253)**

Closes #232.

### Which direction, and why

The issue said either direction is fine — promote it, or remove the need. **Promoted**, for two
reasons:

1. `gan_module` does not merely want a comparison against `build_metadata`'s output. It
   describes the **discriminator** and the **adversarial loss** the same way `build_metadata`
   describes the generator, and a provenance reader must see one spelling of a class path
   across every component of a run. That is a real interface.
2. **This stack already made the identical call one module over** — `_resolved_scale` lost its
   underscore in #249 for exactly this reason. Two private helpers going opposite ways in one
   stack would be the worse outcome.

Being public, it gets a real docstring rather than a one-liner: what it describes (the
instance's *class*, not the instance) and why the shape matters (a recorded value pastes
straight back into a config's own `class_path` key).

### Evidence

- 2 tests proven RED. One is structural: a re-introduced private cross-module import fails it,
  rather than only the current instance being fixed.
- `417 passed` in `tests/training`.
- `mypy` clean; `ruff` clean. No behaviour change.

### DIFF
--- a/sisr/training/gan_module.py
+++ b/sisr/training/gan_module.py
@@ -29,7 +29,7 @@ from ..models.srgan import SRDiscriminator, SRGANEvalConfig, SRGANTrainingConfig
 from ..processors import SRProcessor
 from .config import SREvalConfig
 from .lightning_module import SRLightning
-from .metadata import _class_path, build_metadata
+from .metadata import build_metadata, class_path
 
 #: ``(section, key)`` of every ``build_metadata`` field ``init_from`` validates —
 #: the properties that decide whether one run's generator weights mean anything in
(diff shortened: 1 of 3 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #255, opened 2026-09-01
chore: delete a mutation-testing artifact and cover what it sat inside

### LINKED ISSUE #234: A mutation-testing artifact ships in probe.py
## What is there

`sisr/training/probe.py`, in `_first_probe_dataset`:

```python
    if test_datasets:
        return "test_datasets", next(iter(test_datasets.values()))
    return None  # mutated: unreachable no-op

    return None
```

A dead `return None` after a `return None`, under a comment left over from a mutation experiment.

## Established before proposing to remove it

- It went in with the commit that created the module, not as a deliberate marker.
- It is **not** an anchor the mutation audit expects, so removing it breaks no tooling.
- An AST sweep of `sisr/` finds this is the **only** unreachable statement in the package — so
  this is a one-off, not a pattern.
- `_first_probe_dataset` has no direct test. Worth one while the file is open, since it decides
  the train > val > test priority that makes the contract check fire for a bare `validate` run.

## Acceptance

- [ ] The dead statement and its comment are gone
- [ ] `_first_probe_dataset` gains a direct test of its priority order and its `None` case

YeapJieShen:
**Stack position 10 of 12 · review group: hygiene / UX · base: `refactor/promote-class-path` (#254)**

Closes #234.

```diff
     if test_datasets:
         return "test_datasets", next(iter(test_datasets.values()))
-    return None  # mutated: unreachable no-op
-
     return None
```

Established before removing it, per the issue: it went in with the commit that created the
module rather than as a deliberate marker, it is **not** an anchor the mutation audit expects,
and an AST sweep of `sisr/` found it the **only** unreachable statement in the package.

### The half that is worth more than the deletion

`_first_probe_dataset` had **no direct test**, and it decides the `train > val > test` priority
that makes the contract check fire for a bare `validate`/`test` invocation rather than only for
`fit`. Four tests now pin:

- the priority order, each level in turn
- `None` for "all three unset" **and** for an empty `test_datasets` dict
- that `next(iter(...))` takes the YAML's **first** test set, not an arbitrary one

### The regression guard is a sweep, not an assertion about one line

`test_no_unreachable_statement_survives_in_probe` walks the module's AST for any statement
following a `return`/`raise` in the same block. A re-introduced dead statement **anywhere in
the file** fails it, not just this one. Proven RED.

### Evidence

- `15 passed` in `tests/training/test_probe.py` (11 before).
- `mypy` clean; `ruff` clean.

### DIFF
--- a/sisr/training/probe.py
+++ b/sisr/training/probe.py
@@ -168,8 +168,6 @@ def _first_probe_dataset(dm: Any) -> tuple[str, Any] | None:
     test_datasets = getattr(dm, "test_datasets", None)
     if test_datasets:
         return "test_datasets", next(iter(test_datasets.values()))
-    return None  # mutated: unreachable no-op
-
     return None
 
 
--- a/tests/training/test_probe.py
+++ b/tests/training/test_probe.py
@@ -203,3 +203,80 @@ def test_a_datamodule_with_no_probeable_dataset_yields_no_sample():
     with probe_pair(_DM()) as probe:
         assert probe.sample is None
         assert probe.train_lr() is None
+
+
+# ---------------------------------------------------------------------------
+# _first_probe_dataset: the priority order the contract check depends on
+# ---------------------------------------------------------------------------
+
+
+def test_no_unreachable_statement_survives_in_probe():
+    """A dead `return None` after a `return None`, under a comment left over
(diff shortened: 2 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #256, opened 2026-09-01
fix: test_dataloader raises before setup instead of reporting nothing to do

### LINKED ISSUE #235: test_dataloader() returns an empty list where its siblings raise
## The asymmetry

Verified by constructing an `SRDataModule` and calling each loader without `setup()`:

| call before `setup()` | result |
|---|---|
| `train_dataloader()` | `RuntimeError`, naming the stage that should have run |
| `val_dataloader()` | `RuntimeError`, likewise |
| `test_dataloader()` | **`[]`, silently** |

`_require` exists for exactly this — its message reads *"Lightning calls setup() itself; a direct
caller has to."* The one loader that skips it is the one whose failure is silent.

## Why silent is the wrong failure here

The two states are indistinguishable:

- configured with `Set5`, `setup()` never run -> `[]`
- configured with no test sets at all -> `[]`

A `test` run reaching the first state evaluates nothing and reports success. That is the final
evaluation path, so the number that does not get produced is the one someone would publish.

`predict_dataloader` already shows the fix's shape: it separates "not configured" from "not set
up" and raises with the YAML key to set.

## Also here, latent

`self._test_dl_kwargs is self._val_dl_kwargs` is `True` whenever `test_dataloader_kwargs` is
unset — the same dict object, not a copy. Nothing mutates them today, so it costs nothing now and
couples the two loaders silently the first time anything does. Cheap to fix while in the file.

## Acceptance

- [ ] `test_dataloader()` before `setup()` raises, naming the stage, when test sets are configured
- [ ] It still returns `[]` when no test sets are configured — that is legitimate
- [ ] The kwargs dicts are not aliased
- [ ] Tests for the first two, proven RED

YeapJieShen:
**Stack position 11 of 12 · review group: hygiene / UX · base: `chore/remove-mutation-artifact` (#255)**

Closes #235.

### The asymmetry

| call before `setup()` | before | now |
|---|---|---|
| `train_dataloader()` | `RuntimeError`, naming the stage | unchanged |
| `val_dataloader()` | `RuntimeError`, likewise | unchanged |
| `test_dataloader()` | **`[]`, silently** | `RuntimeError` when test sets are configured |

`_require` exists for exactly this, and its message reads *"Lightning calls setup() itself; a
direct caller has to."* The one loader that skipped it was the one whose failure was silent.

### Why silent is the wrong failure here specifically

The two states were indistinguishable:

```
configured with Set5, setup() never run  ->  []
configured with no test sets at all      ->  []
```

A `test` run reaching the first **evaluates nothing and reports success**. That is the final
evaluation path — the number that does not get produced is the one someone would publish.

`predict_dataloader` already had the fix's shape: separate "not configured" from "not set up".
`test_dataloader` now uses `_test_specs` (what the config asked for) against `_test_ds` (what
`setup` built).

**"No test sets configured" still returns `[]`** and has its own test, so it cannot become an
error by accident.

### Also here, latent

`self._test_dl_kwargs is self._val_dl_kwargs` was `True` whenever `test_dataloader_kwargs` was
unset — the same dict object, not a copy. Nothing mutates them today, so it cost nothing and
would have coupled the two loaders silently the first time anything did. Now a `dict()` copy,
with a test on identity rather than equality.

### Evidence

- 2 tests proven RED; 2 more green on both sides (the legitimate `[]`, and the post-setup path).
- `463 passed` across `tests/training` and `tests/test_cli.py`.
- `mypy` clean; `ruff` clean.

### DIFF
--- a/sisr/training/datamodule.py
+++ b/sisr/training/datamodule.py
@@ -216,7 +216,10 @@ class SRDataModule(lightning.LightningDataModule):
         self._predict_spec = predict_dataset
         self._train_dl_kwargs = train_dataloader_kwargs or {}
         self._val_dl_kwargs = val_dataloader_kwargs or {"batch_size": 1, "num_workers": 1}
-        self._test_dl_kwargs = test_dataloader_kwargs or self._val_dl_kwargs
+        # dict(...), not the object itself: unset, this used to BE
+        # _val_dl_kwargs, so the first thing to mutate either would silently
+        # change both loaders.
+        self._test_dl_kwargs = test_dataloader_kwargs or dict(self._val_dl_kwargs)
         self._predict_dl_kwargs = predict_dataloader_kwargs or {"batch_size": 1, "num_workers": 0}
 
         self._train_ds: Dataset | None = None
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #257, opened 2026-09-01
feat: a progress bar that shows the whole run name and only the wanted metrics

### LINKED ISSUE #237: Progress bar truncates the run name to 4 characters and hardcodes which metrics it shows
## What the bar shows today

```
Epoch 3663:   0%|  | 0/50 [00:00<?, ?it/s, v_num=aled, loss/train/d=0.0217,
loss/train=0.0177, loss/val=0.0049, psnr/val/RGB=24.50, ssim/val/RGB=0.827]
```

Two separate problems, with different causes.

## 1. `v_num=aled` — the version name is truncated to its last four characters

Not ours. Lightning's `get_standard_metrics` does it unconditionally:

```python
if isinstance(version, str):
    # show last 4 places of long version strings
    version = version[-4:]
items_dict["v_num"] = version
```

That rule is written for integer-ish versions (`version_10` -> `_10`). This project names runs
after what they *are*, so a run called `vgg54-scaled` renders as `aled`, and two runs whose names
differ anywhere but the last four characters are indistinguishable in the bar — which is exactly
when you most want to know which one you are watching.

**It cannot be configured away.** The only seam is `ProgressBar.get_metrics()`, and **this project
owns no progress-bar callback** to override it on, so one has to be introduced.

## 2. The metric set is hardcoded at five call sites

Every entry in that bar comes from a `prog_bar=True` that nothing can turn off:

| site | metric |
|---|---|
| `lightning_module.py:647` | `loss/train` |
| `lightning_module.py:707` | `loss/val` |
| `lightning_module.py:720` | first PSNR channel, first SSIM channel |
| `gan_module.py:397` | `loss/train/d` |
| `gan_module.py:420` | `loss/train` (generator total) |

There is no config surface. On an adversarial run that is five metrics plus `v_num`, which is what
wraps the line and pushes the numbers actually being watched — PSNR and SSIM — to the far end.

The existing comment at `lightning_module.py:710` already frames this correctly:

> `prog_bar` is a display choice, so it stays with the caller rather than moving into the scorer

The placement was thought about; the **configurability** was not.

## Proposal

1. **A progress-bar callback owned by this project**, overriding `get_metrics` to render the full
   version name. Nothing else about the bar changes.
2. **A config surface for which metrics appear.** The obvious shape follows `eval_config`'s
   existing vocabulary — a list of families, defaulting to today's behaviour so no shipped config
   changes what it displays, and settable to e.g. `['psnr', 'ssim']` for the requested view.

Where the field lives is the open question. `SREvalConfig` owns what is *scored*, and this is not
that — it is presentation, and the losses in the bar are not eval metrics at all. A field on the
new callback is probably the cleaner home, since a callback is already the thing a YAML wires for
display concerns.

## Risk

**None to any number.** `prog_bar` controls display only; every metric stays logged to TensorBoard
and `metrics.csv` exactly as now. This changes what a human reads mid-run, nothing a result is
computed from.

## Acceptance

- [ ] The full run name appears in the bar, not its last four characters
- [ ] Which metrics appear is configurable, with today's set as the default
- [ ] `['psnr', 'ssim']` yields a bar with only those plus `v_num`
- [ ] No change to what is logged anywhere else — a test that the logged metric set is unaffected

YeapJieShen:
**Stack position 12 of 12 · review group: hygiene / UX · base: `fix/test-dataloader-raises` (#256)**

Closes #237. Last of the stack.

### Before / after

```
before:  v_num=aled, loss/train/d=0.0217, loss/train=0.0177, loss/val=0.0049,
         psnr/val/RGB=24.50, ssim/val/RGB=0.827

after :  v_num=vgg54-scaled, loss/train/d=0.0217, loss/train=0.0177, ...    (default)
after :  v_num=vgg54-scaled, psnr/val/RGB=24.50, ssim/val/RGB=0.827         (metrics: [psnr, ssim])
```

Both verified against a real `TensorBoardLogger`, not only in unit tests.

### 1. The run name

Lightning truncates unconditionally — `version = version[-4:]`, a rule written for integer-ish
versions (`version_10` → `_10`). This project names runs after what they *are*, so
`vgg54-scaled` rendered as `aled` and two runs differing anywhere but their last four
characters were indistinguishable — exactly when you most want to know which you are watching.

### 2. The metric set

`metrics` takes a list of families, defaulting to `None` = **exactly what is shown today**, so
no shipped config changes what it displays.

| setting | bar shows |
|---|---|
| `None` (default) | everything logged `prog_bar=True`, as now |
| `['psnr', 'ssim']` | the scored view |
| `['loss']` | an adversarial run, where PSNR/SSIM get worse by design |

The filter **subtracts as well as selects**, per the request. An unknown family raises rather
than silently rendering an empty bar, which would read as "nothing is being logged".

### The wiring is the subtle part

An unwired callback fixes nothing — Lightning installs its own bar unless something replaces
it. But **`trainer_defaults` is the wrong mechanism**, and I verified why rather than assuming:

```python
# LightningCLI._instantiate_trainer
config[key] += value if isinstance(value, list) else [value]   # APPENDS
```

```
Trainer(callbacks=[SRProgressBar(), TQDMProgressBar()])
-> MisconfigurationException: You added multiple progress bar callbacks
```

So `trainer_defaults` would turn any config naming its own progress bar into a **hard startup
failure** rather than an override. `SRLightningCLI._instantiate_trainer` injects the default
only when no progress bar is already configured.

**A third case was caught by the existing suite, not anticipated by me:**
`enable_progress_bar=False` — `Trainer` refuses a progress-bar callback outright when the bar
is disabled, so three CLI tests failed. Fixed, and it now has its own test.

### One deliberate non-dependency

Lightning's private `_version` helper is **not** imported. Its rule — report a version only
when every logger agrees on one — is reproduced from the public `.version` attribute instead.
A private third-party symbol can vanish in a minor release, and this stack has a PR about
exactly that smell in our own modules.

### Risk

**None to any number.** `prog_bar` controls display only; every metric still reaches
TensorBoard and `metrics.csv` exactly as before.

### Evidence

- 6 tests for the bar + 4 for the wiring, all proven RED first.
- `473 passed` across `tests/training` and `tests/test_cli.py`.
- `mypy` clean; `ruff` clean.

### DIFF
--- a/sisr/cli.py
+++ b/sisr/cli.py
@@ -33,10 +33,12 @@ from typing import Any, Literal
 import torch
 from jsonargparse import set_parsing_settings
 from lightning.pytorch import Trainer
+from lightning.pytorch.callbacks import Callback, ProgressBar
 from lightning.pytorch.cli import ArgsType, LightningArgumentParser, LightningCLI
 
 from .export import to_onnx
 from .training import SRDataModule, SREvalConfig, SRLightning, SRTrainingConfig
+from .training.callbacks import SRProgressBar
 
 # The only keys a checkpoint's hyper_parameters can restore. Legacy checkpoints also
 # carry 'model/*'/'processor'/'criterion', which were TensorBoard-only and never part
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #258, opened 2026-09-01
docs: state that the HR cache checksum keys on name and size only

YeapJieShen:
**Stack position 13 of 13 · base: `feat/progress-bar-callback` (#257)**

No issue — the settled decision from the last review round, written down where the code is.

`compute_checksum` keys on **name and size only**, so replacing an image with a different one of
the same byte length under the same name does not invalidate the cache. Accepted rather than
fixed: keying on `st_mtime_ns` too would make every `git checkout` or `rsync` re-decode the whole
dataset.

Two commits, per the repo's doc-only rule: the docstring, then a test pinning the property so
nobody adds mtime without reading the reason first. The test searches for a same-byte-length PNG
pair rather than assuming one, and skips if it cannot find one, so it cannot flake.

### DIFF
--- a/sisr/datasets/hr_cache.py
+++ b/sisr/datasets/hr_cache.py
@@ -62,6 +62,12 @@ def compute_checksum(img_paths: Sequence[Path]) -> str:
     parameter enters this hash — the cache stores whole raw images, unaffected
     by anything derived from them at read time.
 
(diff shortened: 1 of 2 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #260, opened 2026-09-01
fix: take SRResNet to the reference degradation order; hold SRCNN, measured

### LINKED ISSUE #259: SRResNet degrades an extracted crop; every reference implementation degrades the image and then crops
## The defect

SRResNet's training path bicubic-downsamples an **extracted HR crop**. Every reference
implementation surveyed downsamples the **whole image** and crops paired patches from the
result.

Ours, in `datasets/srresnet.py::TrainDataset.__getitem__`:

```python
hr_crop = <random 96x96 slice of the HR image>
lr_crop = resize(hr_crop, (24, 24))     # degrade the CROP
```

Bicubic support extends past the crop boundary, so degrading a 96x96 crop resolves those taps
against the crop's edge handling instead of against the real neighbouring pixels.

This is the same defect class as the SRCNN one filed separately, on a different architecture,
with a different fix and a different cost. They are filed apart for that reason.

## Measured

96 random scale-aligned crops over 12 DIV2K training images, x4, 96px HR -> 24px LR:

| quantity | value |
|---|---|
| max abs diff over all crops | **22 / 255** |
| whole-patch mean abs diff | 0.156 |
| 2px LR border mean | 0.512 |
| interior mean | **0.000** |
| crops whose interior is bit-identical | **96 / 96** |
| border share of a 24x24 LR patch | **30.6%** |

Every crop's interior is bit-identical and the entire difference is the 2px border — the
signature of the kernel reaching outside the crop and nothing else. Unlike the SRCNN case there
is no grid-shift confound here, because `96 % 4 == 0`.

## Evidence for the convention, stated precisely

**The paper is silent.** Ledig et al. say only:

> "We obtained the LR images by downsampling the HR images (BGR, C=3) using bicubic kernel with
> downsampling factor r=4. For each mini-batch we crop 16 random 96x96 HR sub images of distinct
> training images."

That is suggestive — "the LR images" reads as a full-image artifact — but it does not state the
order, and no code was released. So this is a **convention gap, not a stated-value violation**.

What does settle it is what the field actually does:

- **BasicSR** `extract_subimages.py` crops `DIV2K_train_LR_bicubic/X4` into `X4_sub`. Its input
  is the pre-downscaled **full** LR image; the script contains no resize at all. HR 480 / LR x4
  120, paired.
- **EDSR** `common.get_patch` draws the offset in **LR** coordinates, takes
  `lr[iy:iy+ip, ix:ix+ip]`, then HR at `tx, ty = scale*ix, scale*iy` — LR already exists
  full-size.
- **DIV2K itself** distributes `Train Data Track 1 bicubic downscaling x4 (LR images)` as
  full-image archives. The field's default artifact is a degraded whole image.

## Why it matters here

This is a paper-reproduction framework whose standing rule is that a figure is comparable only to
one computed the same way. The published numbers this project reproduces to ~0.01 dB were
produced by pipelines that all degrade first. Our *training inputs* are therefore not produced
the way the comparators' were, on an axis nobody documented.

Note what is **not** wrong: the degradation kernel is byte-exact MATLAB `imresize` and stays so.
This is about the order the kernel is applied in, not the kernel.

## Consequence the fix carries

Matching the convention forces HR crop offsets to be multiples of `scale`, exactly as EDSR does.
That reduces distinct crop positions 16x at x4 — measured on a 2040x1356 image, 153,892 aligned
positions against 2,452,645 unaligned, still ~123M across the 800-image training set. Accepted,
not overlooked.

## Acceptance

- [ ] SRResNet training pairs are cut from a degraded whole image, matching the surveyed order
- [ ] A test on a crop whose border demonstrably differs today, proven RED first
- [ ] Crop offsets are scale-aligned, and the diversity consequence is stated where it is chosen
- [ ] The whole-image degrade is verified byte-identical against the dataset's own distributed
      LR images, as an external reference

YeapJieShen:
Closes #259. **#245 is closed by #266**, higher in this stack, which lifts the hold this PR places on SRCNN — see "What a real run changed" below for why the hold existed and why it was a sizing requirement rather than a correctness one.

Both train datasets degraded an **already-extracted crop**. Bicubic support
reaches past the crop boundary, so those taps resolved against the crop's edge
handling instead of the real neighbouring pixels. This takes SRResNet to the
reference order and holds SRCNN where it is, for a measured reason.

## What a real run changed about the plan

I first shipped this for both architectures on the strength of a dataset-level
benchmark that reported a 12.2x / 4.8x speedup. **That benchmark was wrong in
direction for SRCNN**, and a real training loop caught it:

| | derived plane | working set | before | after |
|---|---:|---:|---:|---:|
| **SRResNet x4** | 401 MB | 6.7 GB | 22.25 it/s | **23.51 it/s** |
| **SRCNN x3** | 6.3 GB | **12.6 GB** (box has 12 GB) | 21.60 it/s | **1.02 it/s** |

SRCNN is pre-upsampled, so its degraded plane is HR-sized rather than
1/scale^2 of it. Caching it pushes the working set past RAM and a shuffled
loader misses page cache on nearly every item — GPU at 13%, 36% iowait,
992 MB/s sustained read. **So this PR holds SRCNN on per-patch degradation** (it measures 34.23 it/s after
the revert). **#266 lifts that hold**: the constraint is the box's RAM, not the
code, and the box is being given more — shipping the wrong degradation
permanently to fit one machine is the worse trade.

The benchmark's mistake is the reusable part: it walked consecutive indices
(cache-friendly in a way no shuffling loader is) and sampled 2000 of 11 million
patches, so after warmup it measured cache hits rather than the steady-state
miss rate. It now shuffles, and the real-loop probe is what decides.

## Evidence for the order itself

**Primary source, SRCNN.** Dong's released `demo_SR.m` is explicit: `modcrop` ->
`imresize` down -> `imresize` up, then extraction.

**SRResNet: the paper is silent**, so this is a convention gap rather than a
stated-value violation. What settles it is the field — BasicSR's
`extract_subimages.py` crops the already-downscaled `X4` directory and contains
no resize at all; EDSR's `get_patch` draws its offset in LR coordinates and
takes HR at `scale x` that; DIV2K distributes full-image bicubic LR.

## RED -> GREEN

Run against the pre-change modules with the new cache module present, so every
failure is an assertion rather than an import error.

| test | RED on pre-fix |
|---|---|
| SRResNet LR matches the whole-image downscale | `HR offsets must be scale-aligned`, offset (8, 6) |
| SRResNet offsets always scale-aligned | offset (13, 9) is not scale-aligned |

Plus two SRCNN tests that now pin the *deliberate* gap: the LR must equal the
patch's own degradation, **and** must differ from the whole-image order — so if
those ever coincide, the test says the gap is closed rather than passing quietly.

Full suite: **1,082 passed, 2 skipped** (both Windows-only liveness paths).

## Design notes

**A sibling cache, not a second value in the raw-HR one.** That cache's checksum
deliberately ignores every derivation parameter and every shipped config points
every architecture at one directory — a scale-dependent plane living there would
be invalidated by nothing: build at x4, run at x3, same checksum, wrong pixels.
The sibling's name *and* checksum carry kind, scale and an imresize version
marker.

**The marker is bumped by hand, not hashed from source.** A hash would make a
docstring edit trigger a multi-gigabyte rebuild — the same failure mode that made
the raw-HR cache key on name and size only.

**`derived_kind=None` is a documented opt-out**, not an omission: it is how an
architecture whose plane would not fit in RAM declines one.

**Torch-free**, verified by a fresh-interpreter probe: `process_derived_image` is
what gets pickled to build workers, and a spawned worker re-imports its own
defining module.

**Scale-aligned offsets** cost 16x the distinct crop positions at x4 — 153,892
against 2,452,645 on a 2040x1356 image — leaving ~123M across DIV2K-800.
Accepted, and stated where it is chosen.

### DIFF
--- a/sisr/datasets/base.py
+++ b/sisr/datasets/base.py
@@ -23,6 +23,7 @@ import torch
 from PIL import Image
 
 from ..utils.cache import LMDBCache, LMDBCacheBuildContext
+from . import derived_cache
 from .hr_cache import CACHE_NAME, FORMAT_TAG, HEADER, compute_checksum, estimate_map_size
 
 
@@ -116,11 +117,21 @@ class HRCachedTrainDataset(SRDataset):
     SRResNet's train datasets need identically — so the two cannot drift on
     it (e.g. two different build progress-bar labels for what is, by design,
     one shared cache). A subclass supplies only its indexing scheme (grid vs
-    random crop, via :attr:`_img_sizes`) and its read-time degradation, built
-    on :meth:`_read_hr`.
+    random crop, via :attr:`_img_sizes`).
+
+    It owns **two** caches. The raw-HR one is shared verbatim across
+    architectures and scales. The derived-plane one
+    (:mod:`~sisr.datasets.derived_cache`) is per ``(kind, scale)`` and holds the
+    whole image already degraded, so a crop taken from it matches what the
+    reference pipelines produce — degrading the crop instead resolves the
+    bicubic kernel's edge taps against the crop border. Read them with
+    :meth:`_read_hr` and :meth:`_read_derived`.
 
     Args:
         img_dir: Directory of HR images.
+        scale: Upscaling factor the derived plane is built for.
+        derived_kind: Which plane to derive — see
+            :data:`~sisr.datasets.derived_cache.KINDS`.
         use_tqdm: Whether to display a progress bar during the LMDB build.
         cache_dir: Defaults to ``img_dir / '.lmdb_cache'``.
         build_num_workers: ``None`` (default) uses ``min(os.cpu_count() or
@@ -135,6 +146,8 @@ class HRCachedTrainDataset(SRDataset):
     def __init__(
         self,
         img_dir: str | Path,
+        scale: int,
+        derived_kind: str,
         use_tqdm: bool = False,
         cache_dir: str | Path | None = None,
         build_num_workers: int | None = None,
@@ -144,6 +157,7 @@ class HRCachedTrainDataset(SRDataset):
         self._index_images(img_dir)
         self.build_num_workers = build_num_workers
         self._img_sizes = self._collect_sizes()
+        self._derived_kind = derived_kind
 
         cache_dir = Path(cache_dir) if cache_dir else self.img_dir / ".lmdb_cache"
         self._cache = LMDBCache(
@@ -154,6 +168,26 @@ class HRCachedTrainDataset(SRDataset):
             map_size=estimate_map_size(self._img_sizes),
             metadata={"format": FORMAT_TAG},
             build_fn=self._build,
+            use_tqdm=use_tqdm,
+        )
+        # A sibling database in the same directory, not a second value in the one
+        # above: that cache's checksum deliberately ignores every derivation
+        # parameter, so a scale-dependent plane stored there would never be
+        # invalidated by a scale change.
+        self._derived = LMDBCache(
+            cache_dir=cache_dir,
+            name=derived_cache.cache_name(derived_kind, scale),
+            checksum=derived_cache.compute_checksum(self.img_paths, derived_kind, scale),
+            length=len(self.img_paths),
+            map_size=derived_cache.estimate_map_size(self._img_sizes, derived_kind, scale),
+            metadata={"format": derived_cache.cache_name(derived_kind, scale)},
+            build_fn=lambda ctx: ctx.parallel_build(
+                items=self.img_paths,
+                process_fn=derived_cache.process_derived_image,
+                process_args=[(i, derived_kind, scale) for i in range(len(self.img_paths))],
+                num_workers=self.build_num_workers,
+                desc=f"Building {derived_kind} cache (x{scale})",
+            ),
             use_tqdm=use_tqdm,
         )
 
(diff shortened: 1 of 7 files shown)


YeapJieShen:
## Throughput before/after — it is a speedup, not a cost

This issue's scope note assumed matching the reference order would cost
throughput. It does the opposite. The per-item bicubic resize was what remained
after the LMDB cache removed the decode, and slicing a pre-derived plane deletes
it.

Dataset-level `__getitem__` — no model, no GPU, no DataLoader workers, since a
pool would measure the pool. 50 warmup, 400 samples, medians. On mains, GPU
idle, nothing else running, real DIV2K training images.

| dataset | before | after | change |
|---|---:|---:|---|
| SRResNet x4, 96px crop | 10.1086 ms/item — 98.9 items/s | **0.8316 ms/item — 1,202.4 items/s** | **12.2x faster** |
| SRCNN x3, 33px patch | 0.3891 ms/item — 2,569.7 items/s | **0.0810 ms/item — 12,342.8 items/s** | **4.8x faster** |

SRResNet's 10.1 ms/item was a real bottleneck: at batch 16 that is ~162 ms of
data work per batch to hide behind the step.

**What it costs instead is disk, once per (directory, kind, scale).** The LR
plane at x4 is **401 MB**, built in **62.8 s**. SRCNN's pre-upsampled plane is
HR-sized by construction: **6.3 GB**, built in **101.9 s**. The raw-HR cache is
untouched and still shared across architectures.

YeapJieShen:
## Correction — my earlier throughput comment on this issue was measured wrong

**Retracting the "12.2x / 4.8x speedup" I posted above.** It came from a
dataset-level microbenchmark that does not represent a training loader, and for
SRCNN it was wrong in **direction**, not merely magnitude.

### What the benchmark got wrong

- It walked **consecutive** indices. For SRCNN that reads neighbouring patches of
  the same image, which is far more page-cache-friendly than anything a real run
  does — a real loader shuffles across every patch.
- 2000 samples out of 11 million patches touch roughly 13 MB. After warmup it
  was measuring cache hits, not the steady-state miss rate.
- One process, where a real run has 16 workers competing for the same page cache.

### What a real training loop says

400 real training steps, no validation, on mains, GPU otherwise idle:

| | derived plane | working set | before | after |
|---|---:|---:|---:|---:|
| **SRResNet x4** | 401 MB | 6.7 GB | 22.25 it/s | **23.51 it/s** |
| **SRCNN x3** | 6.3 GB | **12.6 GB** | 21.60 it/s | **1.02 it/s** |

**A 21x regression for SRCNN.** The box has 12 GB of RAM. SRCNN is
pre-upsampled, so its degraded plane is HR-sized rather than 1/scale^2 of it;
caching it doubles the working set past RAM and a shuffled loader then misses on
nearly every item. Confirmed at the time by the symptom, not just the rate: GPU
at 13%, 36% iowait, 992 MB/s sustained disk read.

SRResNet is unaffected — its plane is 1/16 the pixels, the working set still
fits, and it gains slightly.

### What changed as a result

SRCNN **keeps degrading each extracted patch**, and this fidelity gap stays open.
SRResNet takes the whole-image order, verified byte-exact against DIV2K's own
distributed LR. `derived_kind=None` is now an explicit, documented opt-out on the
shared base rather than an omission.

After the revert SRCNN measures **34.23 it/s** — above the original 21.60,
because the 6.3 GB plane is no longer competing for page cache.

### What would close this properly

Caching the small LR plane (~0.7 GB at x3) and upsampling only the needed window
at load, with the output grid aligned to the whole-image resize. That keeps the
working set inside RAM and preserves byte-exactness, but it needs an `imresize`
variant that resizes a window at a global grid offset — real work, and it must
re-prove byte-identical output.

Alternatively this is simply a small-box problem: at 16 GB or more the cached
plane fits and the original approach is fine as written.

YeapJieShen (commented):
Human reviewed, all good.

PR #262, opened 2026-09-01
test: verify the degradation order against two external reference sets

YeapJieShen:
Stacked on #260.

`derive` is the single owner of the degradation the training path now applies,
so its convention is checkable against data this project did not produce.

## Result

| reference set | scales | result |
|---|---|---|
| MATLAB-generated benchmark LR (Set5/Set14/B100) | x2, x3, x4 | **357/357 byte-identical** |
| DIV2K's own distributed `LR_bicubic/X4` | x4 | **800/800 byte-identical** |

The second is the one worth the download: it is the actual training set at the
actual training scale, generated independently by the dataset authors. Our
whole-image degradation reproduces their artifact exactly, which turns
"our order matches the field's" from an argument into a measurement.

## Why the existing suite could not already say this

`tests/utils/test_imresize.py` pins the *kernel* against the same benchmark
data, and has since long before this change. It cannot pin the **order**,
because it was written when the training path degraded an extracted patch — so
the only whole-image path it could check was the validation loader's. These
tests close that gap on the training path.

## Skips are not passes

Both legs skip cleanly when their data is absent, so CI stays hermetic. **A skip
means the leg was not exercised** — check the skip count, not just the exit
code. The DIV2K leg needs `DIV2K_train_LR_bicubic_X4.zip` unzipped under
`data/`; the message names it.

The DIV2K leg takes ~3m50s over 800 2K-resolution images. It is data-gated, so
it never runs in CI.

### DIFF
--- a/tests/datasets/test_derived_cache.py
+++ b/tests/datasets/test_derived_cache.py
@@ -128,3 +128,89 @@ def test_process_derived_image_round_trips_through_its_header(image_paths):
     plane = np.frombuffer(value, np.uint8, offset=dc.HEADER.size).reshape(h, w, 3)
     expected = dc.derive(np.array(Image.open(image_paths[1]).convert("RGB")), "lr", 3)
     assert np.array_equal(plane, expected)
+
+
+# ---------------------------------------------------------------------------
+# Byte-equality against externally generated reference data
+#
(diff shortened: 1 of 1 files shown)


YeapJieShen (commented):
Human reviewed, all good.

PR #265, opened 2026-09-01
feat: document per-step scalar logging cost and add an opt-out

### LINKED ISSUE #117: Document the per-step cost of scalar logging and offer an opt-out
## Context

`self.log` is called per training step for several scalars. Its per-step cost is real and, once
the fastest configuration is in play, it becomes the largest remaining host-side term.

`log_every_n_steps` gates the **flush**, not the accumulate, so the existing cadence knob does not
touch this cost.

## Position

Those scalars are the run's evidence. This is a **feature**, not something to silently remove. The
work is to document the cost inline and offer an opt-in "log fewer scalars per step" surface, then
let the user decide per run.

## Acceptance

- [ ] Before/after ms/step at the reference config, measured on this machine
- [ ] An explicit list of which scalars stop existing when the knob is engaged
- [ ] The cost documented at the knob, so the trade is visible where it is made

YeapJieShen:
Closes #117. Stacked on #262.

**Read the measurement before the code — it does not support this issue's
premise, and I have shipped the change anyway with that stated at the knob.**

## What was measured

Real Lightning fit loops, CUDA-synced per step, warmup-excluded medians, on
mains with the GPU otherwise idle.

| configuration | step | `self.log` cost |
|---|---:|---|
| **SRCNN b128, 33x33 Y, scalar MSE** | 6.167 ms | **+1.132 ms — 18.4% of the step** |
| **SRResNet b16, 96x96, composite (mse + tv)** | 36.102 ms | **below noise** |

For SRResNet the three configurations came back at 36.102 / 36.199 / 36.268
ms/step for `all` / `essential` / logging stubbed out entirely — the deltas are
*negative*, i.e. the whole effect is under the noise floor.

The 1.132 ms figure **reproduces the older 1.18 ms measurement** on current
code, which is worth knowing on its own: that figure predated the graph-path
retirement and its stated framing no longer existed.

## Why the knob cannot recover it

The 1.132 ms is the cost of **one** `self.log` call, and on SRCNN that call is
`loss/train` — the objective. `'essential'` keeps it, correctly: this issue's
own position is that these scalars are the run's evidence and must never be
silently removed.

So on both shipped configs the opt-out saves nothing measurable. It pays only
for a criterion with many terms, where the decomposition is several extra calls
per step. That is stated in the docstring, at the knob, rather than left to be
discovered.

## What ships

`training_config.scalar_logging: Literal["all", "essential"] = "all"`.

- **`'all'`** — unchanged. Every existing config behaves exactly as before.
- **`'essential'`** — keeps `loss/train` (and `loss/train/d` under manual
  optimization); drops `loss/train/content`, `loss/train/adv`, and every
  `loss/train/<term>` a composite criterion exposes via `last_terms`.
- Validation is untouched — per-cycle, not per-step.

It lives on the **training config**, not a callback: it changes what a run
*records*, and two runs that logged different scalar sets must not silently look
comparable. The training config reaches `hparams` and the checkpoint; a callback
does not.

## RED -> GREEN

`test_scalar_logging_essential_drops_the_per_step_terms_and_keeps_validation`,
run against the pre-change modules:

```
AssertionError: per-step term tags must be gone under scalar_logging='essential'
assert not ({'loss/train/tv', 'loss/train/vgg22'} & {...})
```

GREEN after. It ships its own mutation — `loss/train` and every validation tag
are asserted still present, so a change that merely stopped logging would fail
here rather than look like a successful opt-out. Plus a field-validation test
(a typo must not silently produce a run with a different evidence set).

Full suite: **1,086 passed, 2 skipped** (both Windows-only liveness paths).

## The call this leaves you

The knob is implemented, tested, default-off and honest about its own value. If
you would rather not carry a config field that recovers nothing on either
shipped architecture, dropping this PR and keeping only the measurement is a
defensible read of the same evidence — the numbers are recorded either way.

### DIFF
--- a/sisr/training/config.py
+++ b/sisr/training/config.py
@@ -102,6 +102,35 @@ class SRTrainingConfig:
             to eager on SRResNet**, so a compiled run does not reproduce an
             eager one. Staying at ``None`` is the standing recommendation.
 
+        scalar_logging: Which per-step scalars a training step records.
+            ``self.log`` costs real host time on **every** step, and
+            ``log_every_n_steps`` gates the *flush*, not the accumulate, so the
+            existing cadence knob does not touch it. Those scalars are the run's
+            evidence, so this is opt-in and never silently applied.
+
+            - ``'all'`` (default) -- unchanged behaviour, every scalar below.
+            - ``'essential'`` -- keeps only the objective each optimiser
+              actually steps on: ``loss/train``, plus ``loss/train/d`` under
+              manual optimization. **Stops recording:** ``loss/train/content``
+              and ``loss/train/adv`` on the adversarial path, and every
+              per-term tag a composite criterion exposes through
+              ``last_terms`` (``loss/train/<term>``). Validation scalars are
+              untouched -- they are per-cycle, not per-step.
+
+            **What this is measured to be worth, so the trade is visible
+            here rather than inferred.** On an RTX 5060 Laptop: one ``self.log``
+            call costs **1.132 ms on a 6.167 ms SRCNN step (18.4%)**, and is
+            **below the noise floor of a 36 ms SRResNet step**. But that 1.132 ms
+            is ``loss/train`` itself — the objective, which ``'essential'``
+            keeps — so on both *shipped* configs this knob recovers nothing
+            measurable. It pays only for a criterion with many terms, where the
+            decomposition is several extra calls per step. Treat it as a lever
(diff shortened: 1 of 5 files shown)


YeapJieShen:
Closing unmerged. The measurement in this PR does not support the change: the opt-out cannot reach the cost that exists, because that cost is `loss/train` itself. Recorded on #117, which closes with it.

PR #266, opened 2026-09-02
fix: take SRCNN to the reference degradation order too

### LINKED ISSUE #245: SRCNN degrades an extracted patch; the authors degrade the image and then extract

YeapJieShen:
Closes #245. Stacked on #262.

Takes SRCNN to the same reference degradation order #260 gave SRResNet, reverting
the hold I placed on it — and documents what that hold was actually about.

## Why the hold is lifted

It was never a correctness objection. The order is right: the authors' released
`demo_SR.m` is explicit (`modcrop` -> `imresize` down -> `imresize` up, *then*
extraction), and degrading an extracted patch resolves the bicubic kernel's edge
taps against the patch border. At x3 the patch interior is bit-identical and the
**entire** difference is a 4px border — 43% of a 33x33 patch.

The objection was **memory**, and memory is a property of the box, not of the
code. SRCNN is pre-upsampled, so its degraded plane is HR-sized rather than
`1/scale**2` of it, and caching it roughly doubles the working set a shuffling
loader touches. Measured on a 12 GB box against a 12.6 GB working set:

| | before | after |
|---|---:|---:|
| real training loop, 400 steps | 21.60 it/s | **1.02 it/s** |

GPU at 13%, 36% iowait, 992 MB/s sustained read. That is a **step change at the
point the working set stops fitting in page cache**, not a gradient — so it is a
sizing requirement, not a reason to ship the wrong degradation.

## 🚨 Read before merging: this needs RAM

**Budget roughly two bytes of RAM per HR pixel in the training set, plus
headroom.** For DIV2K-800 that is ~12.6 GB, so **16 GB is the practical floor and
12 GB is not enough.** On a box below it, SRCNN training will be I/O-bound and
roughly 20x slower, silently — there is no error, only a slow run.

Both `datasets/srcnn.py` and `datasets/derived_cache.py` now state this where the
choice is made, and name the symptom (low GPU utilisation, high iowait, sustained
disk read) because **it cannot be seen in a per-item benchmark** — per-item
*work* is not what changes.

I did not add a runtime memory check. A host-specific guard shipping in a public
framework is something this project has declined before, and the honest version
of it would have to read total RAM and estimate a working set on every platform.
Say the word if you would rather have an opportunistic warning that degrades to
silence where it cannot measure.

## What ships

- SRCNN slices both sub-images out of whole-image planes at the same grid
  position; no kernel reaches past a patch edge.
- `derived_kind` is required again on the shared base — the `None` opt-out had no
  remaining caller, and speculative generality is what the deletion test rejects.
- The `sub_img_size` / `scale` coupling **dissolves**: with no per-patch sampling
  grid left to shift, `33 % scale` stops mattering. There is a test for it across
  x2 / x3 / x4.

## Tests

The whole-image-order tests come back, including the one that ships its own
mutation — it recomputes the *old* per-patch order and fails if the dataset
returns it, so it cannot pass on either implementation.

Full suite: **1,084 passed, 2 skipped** (both Windows-only liveness paths).

### DIFF
--- a/sisr/datasets/base.py
+++ b/sisr/datasets/base.py
@@ -131,11 +131,7 @@ class HRCachedTrainDataset(SRDataset):
         img_dir: Directory of HR images.
         scale: Upscaling factor the derived plane is built for.
         derived_kind: Which plane to derive — see
-            :data:`~sisr.datasets.derived_cache.KINDS` — or ``None`` to build
-            no derived cache at all. ``None`` is not a convenience: a plane
-            that pushes the working set past RAM makes shuffled training
-            I/O-bound, which is measurably far worse than the per-item work
-            it removes. See :mod:`sisr.datasets.srcnn`.
+            :data:`~sisr.datasets.derived_cache.KINDS`.
         use_tqdm: Whether to display a progress bar during the LMDB build.
         cache_dir: Defaults to ``img_dir / '.lmdb_cache'``.
         build_num_workers: ``None`` (default) uses ``min(os.cpu_count() or
@@ -151,7 +147,7 @@ class HRCachedTrainDataset(SRDataset):
         self,
         img_dir: str | Path,
         scale: int,
-        derived_kind: str | None,
+        derived_kind: str,
         use_tqdm: bool = False,
         cache_dir: str | Path | None = None,
         build_num_workers: int | None = None,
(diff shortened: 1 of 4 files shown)


YeapJieShen (commented):
Human reviewed, all good.

